PR #10, opened 2017-02-07
It's time to switch to Morty

sashko:


pelux-base.xml
```diff
@@ -2,25 +2,25 @@
...
-  <project remote="github" revision="master"  name="Pelagicore/meta-pelux"           path="sources/meta-pelux" />
+  <project remote="github" revision="master" name="Pelagicore/meta-pelux"  path="sources/meta-pelux" />
 
   <!-- GENIVI stuff -->
-  <project remote="yocto" revision="11.0" name="meta-ivi" path="sources/meta-ivi" />
+  <project remote="yocto" revision="experimental" name="meta-ivi" path="sources/meta-ivi" />
 
   <!-- Qt Support stuff -->
-  <project remote="github" revision="krogoth" name="meta-qt5/meta-qt5" path="sources/meta-qt5"/>
+  <project remote="github" revision="morty" name="meta-qt5/meta-qt5" path="sources/meta-qt5"/>
```
GordanM: I would prefer to use the meta-qt5 fork from code.qt.io (http://code.qt.io/cgit/yocto/meta-qt5.git/log/?h=5.8). 
So far it has proven to be more clean, stable and the releases follow both the minor and mayor release versions very well. Also, the maintainers push the patches to upstream meta-qt5 and contribute to the -next branches. So the patches you apply on the code.qt.io fork will be applied to mainline once that Qt version is supported by the mainline meta layer.

Gagi2k: +1

DunderRoffe: +1

sashko: Will do with separate commit.

PR #17, opened 2017-03-02
Update README

DunderRoffe:
Update README adding more detailed and up to date instructions for how to build PELUX

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,6 @@ Maintained at https://github.com/pelagicore/pelux-manifests
 Building an Image
 -----------------
 Variables:
-* Branch, what version of this repository to use. This follows the same principles as the Yocto project, see `Branching` below. 
-
 * Manifest, refers to what `<manifest-name>.xml` file you want to use, for example `pelux-intel.xml`. Each hardware platform targeted by the PELUX reference has its own manifest describing what other git repositories are needed for the build.
 
 ### Using vagrant
@@ -19,14 +17,28 @@ Dependencies:
 * Virtualbox
 * virtualization enabled in bios
 
-On Linux systems:
+Procedure:
+
+1. Clone the pelux-manifests git repository.
+2. Start vagrant
 ```bash
-MANIFEST=<manifest> BRANCH=<branch> vagrant up
+MANIFEST=<manifest> vagrant up
 ```
+
+Note:
+The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it
+to set up the build environment. This means that the branch/commit currently checked out will determine what version
+is being built.
 
 ### Using Repo tool
 
+Using this approach one should not clone this git repository. All git manipulation is handled by repo tool.
+Branch denotes what branch of pelux-manifests to use.
+
 ```bash
+mkdir pelux
+cd pelux
+
 repo init -u https://github.com/Pelagicore/pelux-manifests.git -m <manifest> -b <branch>
 repo sync
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 PELUX baseline manifests
 =========================
-This is the manifest repository for building the PELUX reference for various targets using Yocto
+This is the manifest repository for building the PELUX baseline for various hardware targets using Yocto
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,14 @@
-Pelux reference manifests
+PELUX baseline manifests
 =========================
-This is the manifest repository for building the Pelux reference for various targets using Yocto 
+This is the manifest repository for building the PELUX reference for various targets using Yocto
```
breakreturn: Should it say "baseline" here as well?

What "targets" mean here should be made more clear. I assume this means NUC or Minnowboard right now?

README.md
```diff
@@ -1,15 +1,14 @@
...
 
 Maintained at https://github.com/pelagicore/pelux-manifests
 
 Building an Image
 -----------------
 Variables:
-* Manifest, refers to what `<manifest-name>.xml` file you want to use, for example `pelux-intel.xml`. Each hardware platform targeted by the Pelux reference has its own manifest describing what other git repositories are needed for the build.
-* Branch, what version of this repository to use. This follows the same principles as the Yocto project, see `Branching` below. 
-
+* Manifest, refers to what `<manifest-name>.xml` file you want to use, for example `pelux-intel.xml`. Each hardware platform targeted by the PELUX reference has its own manifest describing what other git repositories are needed for the build.
```
breakreturn: "reference" -> "baseline" ?

README.md
```diff
@@ -1,15 +1,14 @@
...
+* Image, refers to what version of PELUX that should be built. Currently there are two versions: `core-image-pelux` and `core-image-pelux-qt`. The latter being a version that includes NeptuneUI and QtApplicationManager.
```
breakreturn: I believe we should call this "core-image-pelux-qtas", since the point is to have QtAS in there and not "just" Qt.

DunderRoffe: You might be right. However, it is slightly out of scope for this PR. Would you be OK with doing that change in another PR?

README.md
DunderRoffe: Would "hardware targets" be a better than "targets"?

README.md
breakreturn: Yes.

README.md
```diff
@@ -19,14 +17,28 @@ Dependencies:
...
 ```
 
+Note:
+The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it
+to set up the build environment. This means that the branch/commit currently checked out will determine what version
+is being built.
+
 ### Using Repo tool
 
+Using this approach one should not clone this git repository. All git manipulation is handled by repo tool.
```
GordanM: Nitpick: ... by the repo tool.

README.md
```diff
@@ -19,14 +17,28 @@ Dependencies:
...
+Branch denotes what branch of pelux-manifests to use.
```
GordanM: Nitpick: Branch, ... 
For consistency.

README.md
```diff
@@ -19,14 +17,28 @@ Dependencies:
...
+
+1. Clone the pelux-manifests git repository.
+2. Start vagrant
 ```bash
-MANIFEST=<manifest> BRANCH=<branch> vagrant up
+MANIFEST=<manifest> vagrant up
 ```
 
+Note:
+The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it
```
GordanM: Sorry if I've got this wrong but shouldn't the manifest here be in singular form? Or did I misunderstand something.

README.md
jeremiah: Looks good to me.

README.md
DunderRoffe: Since there are multiple manifests to choose from in the `pelux-manifests` repo i think it is fair to say manifests. I guess only one is pointed to but they are linked and if you point to `pelux-intel-qt.xml` that would import all of the other manifests as well.

README.md
DunderRoffe: Created issue for this. See #18

README.md
GordanM: Perfect! Thanks for explaining this!

PR #30, opened 2017-03-23
pelux-base: bump meta-ivi to 12.0

sashko:


pelux-base.xml
```diff
@@ -38,7 +38,7 @@
 
   <!-- GENIVI stuff -->
   <project remote="github"
-           revision="f4ee00a42565ed6640493b7f8c9f1069e9d4b693"
+           revision="12.0"
```
DunderRoffe: Please use static revisions. Tracking branches tends to result in builds failing without anyone doing any updates. This has already happened once to pelux-manifests and it was a real pain to debug.

PR #35, opened 2017-05-03
Update to morty

erikboto:
This updates most layers to use a recent commit from morty-branches.


DunderRoffe:
The build failure is false. I have verified that it actually was successful.

PR #37, opened 2017-05-04
scripts: add mkefiWrap script

DunderRoffe:
Add script for speeding up the process of putting images on SD-cards and
USB-sticks.

Signed-off-by: Viktor Sjölind <viktor.sjolind@pelagicore.com>

### DIFF
--- a/pelux-base.xml
+++ b/pelux-base.xml
@@ -32,7 +32,7 @@
            path="sources/meta-virtualization"/>
 
   <project remote="github"
-           revision="515f5cc2ef3911d9cf471dc4e8ece1b76a34b4c6"
+           revision="d224ee09f62b84be734be6e3959abf48347dfa91"
            name="Pelagicore/meta-pelux"
            path="sources/meta-pelux" />
 


DunderRoffe:
Offline comments from @rpannek  and @erikboto suggests that the scripts should be moved to meta-pelux or meta-bistro. However, I forgot the reasoning could you refresh my memory?

rpannek:
meta-bistro is a layer whith all kinds of software on top of meta-ivi, so it's more probabry to be used more often than the pelux-manifests, especially the manifests will be forked by each project so we can't update this script as easily. But now I'm not sure about the future role of meta-bistro, so meta-plux might be a good way too, or perhaps upstream it to poky and make the functionality part of mkefidisk.sh instead?

erikboto:
When working you usually don't have the manifest git easily accessible since it is only used by repo, but meta-pelux would be easily accessible under sources/meta-pelux.

DunderRoffe:
OK I will move the script to meta-pelux and update this PR accordingly.

DunderRoffe:
Pending: https://github.com/Pelagicore/meta-pelux/pull/10

erikboto:
Now merged in meta-pelux, closing this PR.

PR #42, opened 2017-05-16
meta-pelux: revision update

GordanM:
Adds GENIVI/dlt-daemon and depends on meta-ivi update.

Signed-of-by: Gordan Markuš <gordan.markus@pelagicore.com>


GordanM:
@DunderRoffe I've made a test build which passed but I haven't deployed it on the target.

Issue #22, opened 2017-03-08
Vagrantfile only allows Intel conf


thenor:
When Vagrantfile is setting TEMPLATECONF it checks for a directory which only exists if one of the Intel manifests is used. This is not a problem as long as only Intel manifests exist, but can become an issue in the future.

PR #43, opened 2017-05-17
WIP: Vagrantfile: add possibility to use TEMPLATECONF from other BSP layers

erikboto:
This fixes https://github.com/Pelagicore/pelux-manifests/issues/22. 

It will now try to get TEMPLATECONF from meta-pelux-bsp-<BSP> where BSP is set as an environment variable in the same way MANIFEST and BITBAKE_IMAGE are passed. I made BSP default to intel if nothing is set to not break any previous behavior, but maybe we want to force this input?

Instead of specifying BSP we could also enforce a strict naming scheme of manifests, so that the bsp-name can be extracted from there. Like pelux-\<BSP\>-whatever, but that might just be confusing for future users?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,9 @@ Building an Image
 
 The following manifests can be used for a build:
 
-* `pelux-intel.xml` - For building the `core-image-pelux` image
+* `pelux-intel.xml` - For building the `core-image-pelux` image for Intel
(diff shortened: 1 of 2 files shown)


erikboto:
@DunderRoffe @thenor I've updated now to the other solution, where BSP is extracted from manifest name. Please re-review!

PR #45, opened 2017-05-26
pelux: update components

GordanM:
* Cleanup commits in meta-bistro
* Moving fstab file from the BSP layers to meta-pelux


GordanM:
@DunderRoffe you are right, sure will do!

PR #46, opened 2017-05-30
Jenkinsfile: Switched to pipeline syntax

breakreturn:


rpannek:
@breakreturn you added the invalid label here, should we close this PR if it's invalid?

breakreturn:
@rpannek No I'll close it or remove the label in a bit. I use this PR at the moment for testing the CI flow on pull requests. I just added the label to people wouldn't spend time reviewing some temporary stuff.

DunderRoffe:
A side note is that it would be nice to add an additional step that exports the artifacts from the build. Currently all of them are thrown away.

PR #47, opened 2017-05-31
Support for docker

DunderRoffe:
Add support for running docker as vagrant provider

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,22 @@
+FROM ubuntu:16.04
+
+RUN mkdir /var/run/sshd
+RUN apt-get update
+RUN apt-get install -y openssh-server sudo
+
+# Required by bitbake
+RUN apt-get install -y cpio
+
+# en_US.utf8 is required by Yocot sanity check
+RUN apt-get install -y locales && rm -rf /var/lib/apt/lists/* \
+    && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+ENV LANG en_US.utf8
+
+RUN useradd vagrant --create-home --user-group --groups sudo
+RUN echo 'vagrant:vagrant' | chpasswd
+
+RUN echo 'vagrant ALL=(ALL) NOPASSWD:SETENV: ALL' > /etc/sudoers.d/vagrant
+
+EXPOSE 22
+CMD  ["/usr/sbin/sshd", "-D"]
+
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,11 +26,25 @@ end
 
 Vagrant.configure(2) do |config|
 
-    config.vm.box = "debian/jessie64"
+    # Prefer docker over virtualbox since it is listed first
+    # meaning that docker should run if no '--provider' is supplied
+    # when calling vagrant up
+    config.vm.provider "docker" do |d, configOverride|
+        d.build_dir = "."
+        d.pull = true
+        d.has_ssh = true
 
-    config.vm.provider "virtualbox" do |vb|
+        # Overrides for 'config' unique for docker
+        configOverride.ssh.username = "vagrant"
+        configOverride.ssh.password = "vagrant"
+    end
+
+    config.vm.provider "virtualbox" do |vb, configOverride|
         vb.memory = ram * 1024
         vb.cpus = cpus
+
+        # Overrides for 'configs' unique for virtualbox
+        configOverride.vm.box = "debian/jessie64"
     end
 
 
@@ -97,6 +111,7 @@ Vagrant.configure(2) do |config|
     config.vm.provision "shell",
         args: ["/home/vagrant/pelux_yocto/", bitbake_image],
         privileged: false,
+        env: {"LANG" => "C.UTF-8"},
         path: "vagrant-cookbook/yocto/build-images.sh"
 end
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -33,9 +33,6 @@ Vagrant.configure(2) do |config|
         vb.cpus = cpus
     end
 
-    # Attach an extra disk to /home/vagrant
-    eval (File.read 'vagrant-cookbook/host-system-config/attach-extra-disk.vagrantfile')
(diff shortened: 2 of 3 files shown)


Vagrantfile
```diff
@@ -33,9 +33,6 @@ Vagrant.configure(2) do |config|
         vb.cpus = cpus
     end
 
-    # Attach an extra disk to /home/vagrant
-    eval (File.read 'vagrant-cookbook/host-system-config/attach-extra-disk.vagrantfile')
-    config.vm.provision "shell", args: ["/dev/sdb"], path: "vagrant-cookbook/host-system-config/create-disk.sh"
```
rpannek: Will the available disk have enough space for everything?

DunderRoffe:
Currently I do not get bitbake to run in Docker due to Python complaining about no UTF-8 locale. However, I tried installing it through the `Dockerfile` but still no success. Any ideas?

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+  && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
+ENV LANG en_US.utf8
+
+RUN useradd vagrant --create-home --user-group --groups sudo
+RUN echo 'vagrant:vagrant' | chpasswd
+
+RUN echo 'vagrant ALL=(ALL) NOPASSWD:SETENV: ALL' > /etc/sudoers.d/vagrant
+
+EXPOSE 22
+CMD ["/usr/sbin/sshd", "-D"]
```
rpannek: Shouldn't you do that via systemd instead? Especially the next time you start a docker container you want it running too don't you? Or if you run everything via the docker command you don't need sshd there, getting the artifacts could work by bindmounting in a directory from the outside.

breakreturn:
@dunderroffe Have you tried a less minimal distro to resolve the locale thing?

Vagrantfile
```diff
@@ -26,28 +26,36 @@ end
...
+        d.pull = true
+        d.has_ssh = true
+
+        # Overrides for 'config' unique for docker
+        configOverride.ssh.username = "vagrant"
+        configOverride.ssh.password = "vagrant"
+    end
 
-    config.vm.provider "virtualbox" do |vb|
+    config.vm.provider "virtualbox" do |vb, configOverride|
```
breakreturn: If we get Docker working as we want, I don't think we should support VBox anymore. If we won't use VBox ourselves it's very likely that it will break or become unstable etc, without us noticing it.

jonte: By removing VBox support, we also lose Mac & Windows users (which may not be interesting anyway). Just saying.

rpannek: You can use Docker on Windows https://docs.docker.com/docker-for-windows/ and macOS https://docs.docker.com/docker-for-mac/ too.

jonte: @rpannek .. which will load up a virtual machine.. I guess that will work, not sure if it is better than using VBox directly though.

breakreturn: @jonte Good point. But I'm not saying that it's not wanted, I'm just saying that our support for it is likely to break unless we use it ourselves. Or if we actively test it etc ofc. But I'm not sure that would be prioritized unless we have actual users that wants it. Maybe we do?

jonte: It shouldn't be that work work to just add it back when/if we need it in the future I guess. We just have to be careful to not use docker-specific functionality. If we keep it, we could just have one CI job running on VBox just to make sure it still works. I think having Windows & Mac support is a pretty nice feature.

Vagrantfile
DunderRoffe: I do not know. It might be the case that we need to add this again. However, it does not work in this setup since it is hard coded for VBox. So at the very least it needs some overhaul either way.

Dockerfile
DunderRoffe: Yes you are right. Will fix

Vagrantfile
rpannek: I do think that it is better because most of the users wouldn't have the huge overhead of VirtualBox but those which can't run in nativelly on linux are still able to use it.

breakreturn: @jonte I've started to try and find out internally about how prepared we are to support this, let's see what comes up. In the meantime, we really would benefit from having a Docker provider so for me it's fine to merge this as it is and perhaps change later.

DunderRoffe:
@rpannek tried to change the way sshd is started. Did not work. Lets get a working version first then we can optimize it :)

Vagrantfile
gillesfernandez: I'd say we should go to Docker as it will also provide the support for VBox for us. It's not a lot of work, but still.

DunderRoffe:
CI result
`core-image-pelux` builds fine
`core-image-pelux-qt` fails while building `qtapplicationmanager` and `qtlocation`

```
==> default: NOTE: Tasks Summary: Attempted 4561 tasks of which 19 didn't need to be rerun and 2 failed.
==> default: 
==> default: Summary: 2 tasks failed:
==> default:   /home/vagrant/pelux_yocto/sources/meta-boot2qt/recipes-qt/automotive/qtapplicationmanager_git.bb:do_compile
==> default:   /home/vagrant/pelux_yocto/sources/meta-qt5/recipes-qt/qt5/qtlocation_git.bb:do_compile
==> default: Summary: There were 11 WARNING messages shown.
==> default: Summary: There were 4 ERROR messages shown, returning a non-zero exit code.
```

DunderRoffe:
Solved the build issue by consulting @e8johan which had some advice earlier from @erikboto, long story short: yocto fails to build the qtdbustocpp natively on Debian. However, this is not a problem in Ubuntu. Therefore the docker image now run Ubuntu. Personally I would prefer to run Debian but lets deal with the discussion of what OS to run the builds on outside this PR.

@breakreturn @rpannek @gillesfernandez @jonte what do you think about this solution?

rpannek:
That sounds reasonable to me.

breakreturn:
@DunderRoffe Yeah I agree, let's get this working in the CI system and then keep talking about what we'd like to improve as we go along.

PR #48, opened 2017-06-01
pelux distro: update bistro, PELUX and BSP layers

GordanM:
This introduces PELUX as a Yocto distribution. The distribution conf file
as well as the image files can be found in the meta-pelux layer. Distro
and image configuration has been moved from the BSP meta layers
local.conf files to the meta-pelux layer. The meta-bistro layer is used
as an aggregate of Pelagicore components and bbappends.

### DIFF
--- a/pelux-base.xml
+++ b/pelux-base.xml
@@ -22,7 +22,7 @@
            path="sources/poky"/>
 
   <project remote="github"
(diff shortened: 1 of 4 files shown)


GordanM:
@rpannek done :)

GordanM:
@rpannek can you check the commit message now?

PR #49, opened 2017-06-08
Create manifests directory to clean up repo root

DunderRoffe:
Create a directory called "manifests" where all manifest .xml files are
moved. This is important since the number of manifests is getting big
and the root directory of the repository is starting to get cluttered.

Signed-off-by: Viktor Sjölind <viktor.sjolind@pelagicore.com>


README.md
```diff
@@ -56,7 +56,7 @@ cd pelux
 Instruct repo tool to fetch a manifest using the command `repo init`. In this context, branch denotes what
 branch of git repo `pelux-manifests` to use. Then make repo tool fetch all sources using the command `repo sync`.
 ```bash
-repo init -u https://github.com/Pelagicore/pelux-manifests.git -m <manifest> -b <branch>
+repo init -u https://github.com/Pelagicore/pelux-manifests.git -m manifests/<manifest> -b <branch>
```
breakreturn: This change will need to be reflected in the SWF as well. Might be more than one place but at least there's this: https://pelagicore.github.io/software-factory/chapters/baseplatform/index.html#using-the-repo-tool

rpannek:
For me what makes it cluttered are not the manifests, because they are the main content of this repo but all those special files which just seem to get more and more without the possibility to move them, now we have Jenkinsfile, Dockerfile, Vagrantfile.

None of them have anything to do with the content of this repo, all of them are just there because we chose some tools which want them there, I'm afraid that this will get out of hand soon. If a project decides to use different tools those files will confuse everyone. I really do dislike this trend of forcing tool configuration into the root of a project.

breakreturn:
@rpannek So do you have a proposal of what to do instead or is that just a general comment about these tools and trends etc.?

rpannek:
I guess it was more like a general comment about my disapointment of these trends. But I actually think most of those files don't need to be in the root, Jenkins master has a commit in it which makes it possible to move the Jenkinsfile, Dockerfile doesn't need to bi in the root already and neither does Vagrantfile. Perhaps we should try to move those files instead?

breakreturn:
@rpannek @DunderRoffe I personally think it would be OK to leave the structure as it is if there are changes in the tools coming in that would allow us to move the "special" files instead of the manifests.

GordanM:
Agree with @breakreturn .

DunderRoffe:
I also agree. Closing this  PR

PR #50, opened 2017-06-15
Update readme

### LINKED ISSUE #31: Add documentation on difference in build alternatives
As a new user I want to be the README to cover when it is best to build with vagrant and when it is best to build "manually" with repo-tool. Both have their use but the advantages and disadvantages are not listed.

DunderRoffe:
The README needs to be updated with info on how to build with Vagrant and that it is now possible to use Docker. This also fixed #31.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,12 @@ vagrant ssh -c \"/vagrant/vagrant-cookbook/yocto/build-images.sh
                  ${bitbake_image}
                \"
 ```
+6. Move the built images to the host
+```bash
+vagrant ssh -c "cp -r ${yoctoDir}/build/tmp/deploy/images /vagrant"
+```
 
-The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built.
+The container/virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built. The final step will copy the image directory containing the built images to the directory on the host where vagrant was started.
 
 ### Using Repo tool
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 PELUX baseline manifests
 =========================
-This is the manifest repository for building the PELUX baseline for various hardware targets using Yocto
+This is the manifest repository for building the PELUX baseline for various
+hardware targets using Yocto
 
 Maintained at https://github.com/pelagicore/pelux-manifests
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,8 +13,8 @@ The following manifests can be used for a build:
 * `pelux-intel-qt.xml` - For building the `core-image-pelux-qt` image, which is the baseline with QtAS
 * `pelux-rpi.xml` - For building the `core-image-pelux`image for Raspberry Pi 3
 
-An image build can be started using a virtual machine, see section "Using vagrant", or using `repo` tool
-directly, see section "Using Repo tool".
+An image build can be started using a container/virtual machine, see section "Using vagrant", or using `repo` tool directly, see section "Using Repo tool". Since the Repo tool approach is not boxed it is notably faster than using Vagrant. Therefore, the Vagrant approach is usually only used in situations where one does not want to depend on the host system, such as running continuous integration jobs.
```
rpannek: With the docker image this is not quite true anymore, there it is basically as fast as the native one, the only thing is that it's a bit more complicated to work with because you need to ssh into the docker image and stuff like that.

README.md
```diff
@@ -26,20 +26,38 @@ Variables:
...
+vagrant ssh -c \"/vagrant/vagrant-cookbook/yocto/build-images.sh
+                 ${yoctoDir}
+                 ${bitbake_image}
+               \"
 ```
 
-The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it
-to set up the build environment. This means that the branch/commit currently checked out will determine what version
-is being built.
+The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built.
```
rpannek: Where does one find the artifacts in this case?

README.md
DunderRoffe: Well I guess that is up to the configuration of docker right? I would at least assume that out of the box docker wont have access to all RAM and CPU on the host?

README.md
DunderRoffe: Good point! It is still in the docker so one would have to move them out... Will add another step of how to move it to the host system.

README.md
```diff
@@ -56,8 +56,12 @@ vagrant ssh -c \"/vagrant/vagrant-cookbook/yocto/build-images.sh
                  ${bitbake_image}
                \"
 ```
+6. Move the built images to the host
+```bash
+vagrant ssh -c "cp -r ${yoctoDir}/build/tmp/deploy/images /vagrant"
+```
 
-The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built.
+The container/virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built. The final step will copy the image directory containing the built images to the directory on the host where vagrant was started.
```
rpannek: The added text doesn't do what the commitmessage promisses.

DunderRoffe: OK I do not see it though. How would you have it? :)

rpannek: A now I see the `cp` step which I didn't quite get that it moves the image files to the host. I guess it's a bit unintuitive for someone who never used vagrant.

README.md
```diff
@@ -26,20 +26,42 @@ Variables:
...
+```
+6. Move the built images to the host
+```bash
+vagrant ssh -c "cp -r ${yoctoDir}/build/tmp/deploy/images /vagrant"
 ```
 
-The virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it
-to set up the build environment. This means that the branch/commit currently checked out will determine what version
-is being built.
+The container/virtual machine started via vagrant will sync the cloned git repository and use the manifests contained in it to set up the build environment. This means that the branch/commit currently checked out will determine what version is being built. The final step will copy the image directory containing the built images to the directory on the host where vagrant was started.
```
breakreturn: I don't know if anyone thought about the pros and cons of breaking lines. But in most of our markdown docs, we manually insert linebreaks. I think we should be consistent with that whatever way we choose to do it.

DunderRoffe: I agree

PR #53, opened 2017-06-21
Make baseline and qt jobs possible to run parallel

rpannek:
This change makes it possible to compile the different manifests
on different nodes with Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,66 +5,53 @@ def buildManifest = {String manifest, String bitbake_image ->
     // Store the directory we are executed in as our workspace.
     String workspace = pwd()
 
-    try {
-        // Stages are subtasks that will be shown as subsections of the finished build in Jenkins.
+    // Stages are subtasks that will be shown as subsections of the finished build in Jenkins.
 
-        stage("Checkout ${bitbake_image}") {
-            // Checkout the git repository and refspec pointed to by jenkins
-            checkout scm
-            // Update the submodules in the repository.
-            sh 'git submodule update --init'
+    stage("Checkout ${bitbake_image}") {
+        // Checkout the git repository and refspec pointed to by jenkins
+        checkout scm
+        // Update the submodules in the repository.
+        sh 'git submodule update --init'
+    }
+
+    stage("Start Vagrant ${bitbake_image}") {
+        // Calculate available amount of RAM
+        String gigsramStr = sh (
+            script: 'free -tg | tail -n1 | awk \'{ print $2 }\'',
+            returnStdout: true
+        )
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -72,10 +72,8 @@ def buildManifest = {String manifest, String bitbake_image ->
     sh "vagrant destroy -f || true"
 }
 
-node("DockerCI") {
-    buildManifest("pelux-intel.xml", "core-image-pelux")
-}
+def builders = [:]
+builders['core'] = { node("DockerCI") { buildManifest("pelux-intel.xml", "core-image-pelux") }}
+builders['qt']   = { node("DockerCI") { buildManifest("pelux-intel-qt.xml", "core-image-pelux-qt") }}
```
GordanM: I am not sure if the "qt" builder name makes it obvious that in fact it will include the core image, plus Qt, plus Qt Auto. The difference in `core-image-pelux` and `core-image-pelux-qt` is actually not Qt but Qt Auto. I would suggest the builder names to be changed to something more sensible, do you agree with this @rpannek ?

rpannek: Ah yeah sure, I can change it to qtauto, didn't think of it.

DunderRoffe: It might also be nice to add a comment describing what the parallel command does. Unfortunately, I do not think groovy syntax speaks for it self.

rpannek:
Finally I got the whole parallel thing working, would you mind checking again @GordanM and @breakreturn?

Vagrantfile
```diff
@@ -41,11 +41,16 @@ Vagrant.configure(2) do |config|
     # Install dependencies for BitBake
     config.vm.provision "shell", path: "vagrant-cookbook/deps/yocto.sh"
 
+    # Set UTF-8 for BitBake python
+    config.vm.provision "shell", privileged: false, inline: <<-SHELL
```
GordanM: Nice

Vagrantfile
```diff
@@ -41,11 +41,16 @@ Vagrant.configure(2) do |config|
...
+        echo 'export LC_ALL="en_US.UTF-8"' >> .profile
+    SHELL
+
     # Configure username and password in git
     config.vm.provision "shell", privileged: false, path: "vagrant-cookbook/system-config/vagrant-ssh-user.sh"
 
     # Configure username and password in git
     config.vm.provision "shell", privileged: false, path: "vagrant-cookbook/yocto/initialize-repo-tool.sh"
-
+
```
GordanM: Super ultra nitpick, could you please remove the dangling space?

rpannek:
Done.

PR #54, opened 2017-06-30
Add archive step to Jenkinsfile

rpannek:


Jenkinsfile
```diff
@@ -63,5 +76,5 @@ def buildManifest = {String manifest, String bitbake_image ->
 parallel 'core':{
     node("DockerCI") { buildManifest("pelux-intel.xml", "core-image-pelux") }
 },'qtauto':{
-    node("DockerCI") { buildManifest("pelux-intel-qt.xml", "core-image-pelux-qt") }
+    //node("DockerCI") { buildManifest("pelux-intel-qt.xml", "core-image-pelux-qt") }
```
GordanM: Looks good but it seems like you have commented out a line here. Is there a reason behind this?

rpannek:
@GordanM I rewrote it to only add the archive step, I will work on the mirror stuff in a different PR, could you re-review it now?

Jenkinsfile
```diff
@@ -54,6 +54,18 @@ def buildManifest = {String manifest, String bitbake_image ->
         sh "vagrant ssh -c \"/vagrant/vagrant-cookbook/yocto/build-images.sh ${yoctoDir} ${bitbake_image}\""
     }
 
+    stage("Copy images and cache ${bitbake_image}") {
+        sh "rm -rf archive"
+        sh "mkdir -p archive"
+        sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/tmp/deploy/images/ /vagrant/archive/\""
+        sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/downloads/ /vagrant/archive/\""
```
DunderRoffe: Is it really necessary to copy the downloads? To my mind they are not needed because you have what you need in the sstate-cache or you want to build something that is not in the sstate-cache and then you would not find it in downloads either.

DunderRoffe:
CI build fails due to:
```
[core] OSError: [Errno 28] No space left on device
.
.
.
[qtauto] OSError: [Errno 28] No space left on device
```

Jenkinsfile
rpannek: For some reason docker seems to be the default provider, both on my computer and on the build server:

    Bringing machine 'default' up with 'docker' provider...

rpannek:
Because the build error is unrelated to che changes and I have build it a couple of times successfull already I'm merging this change now.

Issue #18, opened 2017-03-03
Rename "core-image-pelux-qt" to "core-image-pelux-qtas"


DunderRoffe:
Broken out from a discussion in #17 

@breakreturn:
I believe we should call this "core-image-pelux-qtas", since the point is to have QtAS in there and not "just" Qt.

PR #55, opened 2017-06-30
Rename core-image-pelux-qt to core-image-pelux-qtauto

GordanM:
This was done in order to explicitly reflect that the qtauto image adds
Qt Automotive Suite components.


GordanM:
Fixes #18 .

PR #60, opened 2017-07-27
Use ARCHIVE_CACHE env variable

rpannek:
Most of the time we don't need to archive the sstate and download
cache. For the moments we do, this patch offers a environment variable
which one can set via a parameterized Jenkins job.

Mostly this would be used by the nightly jobs to freshen up the
shared caches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,8 +61,9 @@ def buildManifest = {String manifest, String bitbake_image ->
         sh "mkdir -p archive"
         sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/tmp/deploy/images/ /vagrant/archive/\""
 
-        // Archive the downloads and sstate when we build the master
-        if (env.BRANCH_NAME == "master") {
+        // Archive the downloads and sstate when the environment variable was set to true
+        // by the Jenkins job.
+        if (env.ARCHIVE_CACHE) {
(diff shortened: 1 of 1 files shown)


DunderRoffe:
@rpannek what does ARCHIVE_CACHE do?

DunderRoffe:
Can not find any documentation on it

rpannek:
It's just a environment variable which you can set from Jenkins for example, I made it bool so I can make a parametrized job and tell the Jenkinsfile to archive the cache.

Jenkinsfile
```diff
@@ -61,8 +61,8 @@ def buildManifest = {String manifest, String bitbake_image ->
         sh "mkdir -p archive"
         sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/tmp/deploy/images/ /vagrant/archive/\""
 
-        // Archive the downloads and sstate when we build the master
-        if (env.BRANCH_NAME == "master") {
+        // Archive the downloads and sstate when we build the master in multi pipeline
+        if (env.ARCHIVE_CACHE || env.BRANCH_NAME == "master") {
```
rpannek: Hm, or perhaps we shouldn't archive the cache on a multi pipeline job, just during the nightly builds which can be parametrized?

rpannek:
Yeah, I measured it and just the archiving of the new cache takes about 30 minutes because it needs to be first moved between servers, then copied between the docker image and where it can be served by Nginx. It's a lot of data. I guess we will need more measurements later to see which one makes the whole process the fastest.

PR #62, opened 2017-07-27
Use a site.conf for incremental builds unless building master

rpannek:
Needs global Jenkins environmental variable YOCTO_CACHE_URL
to be set to work, which can be added under:

Manage Jenkins -> Configure System -> Global Properties


DunderRoffe:
@rpannek The shortlog of the commit message is wrong now since we do not care about `BRANCH_NAME=="master"` anymore

rpannek:
Ok you're right, I will change that.

PR #65, opened 2017-09-01
README.md: Editing for clarity

jeremiah:
Changes to make building an image clearer (hopefully.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ determine what version is being built. The final step will copy the image
 directory containing the built images to the directory on the host where vagrant
 was started.
 
-### Using Repo tool
+### Using the "repo" tool
 
 #### NOTE: When using this approach one should not clone this git repository, all git manipulation is handled by the repo tool.
 
@@ -130,6 +130,8 @@ information on the different images.
 ```bash
 bitbake <image>
 ```
+## Test image
+The next step is to test your image if your build was successful.
 
 Targets
 -------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,7 +1,7 @@
 PELUX baseline manifests
 =========================
-This is the manifest repository for building the PELUX baseline for various
-hardware targets using Yocto
+This manifest repository is used for building the PELUX baseline for various
+hardware targets using the Yocto build tool.
```
breakreturn: I don't feel it's more correct like this Yocto is usually referred to as more than the tool used for building. Maybe "using the Yocto project"?

README.md
```diff
@@ -112,13 +112,13 @@ repo init -u https://github.com/Pelagicore/pelux-manifests.git -m <manifest> -b
 repo sync
 ```
 
-When done fetching the sources, create a build directory and set up bitbake.
-TEMPLATECONF tells the `oe-init-build-env` script which path to fetch
-configuration samples from. Note that the example below get the template
-configuration for the Intel BSP, adapt the path according to your current BSP.
+When 'repo sync' has finished fetching the sources, the next step is to create a 'build' directory and set up bitbake.
```
DunderRoffe: I would prefer ` (back ticks) over ' which puts the content in a box making it clearer that you are talking about a command.

README.md
GordanM: +1

README.md
```diff
@@ -112,13 +112,13 @@ repo init -u https://github.com/Pelagicore/pelux-manifests.git -m <manifest> -b
...
+The TEMPLATECONF envorinment setting tells the `oe-init-build-env` script which path to fetch
```
tobsan: environment*

README.md
```diff
@@ -138,14 +140,12 @@ information about how to use a built image with the targets, see
 [Getting started](getting-started.md).
 
 ### Pelux Intel
-Reference instance for the Intel i7 x86 platform. Examples of boards using this
-architecture include:
+Reference instance for the Intel i7 x86 platform. 
 
-* Intel NUC
-* Minnowboard
+* [Intel NUC](https://en.wikipedia.org/wiki/Next_Unit_of_Computing)
```
tobsan: Please keep the Minnowboard here, as it, from a Yocto perspective, uses the i7 platform. Perhaps we can change the wording to just say generic Intel amd64 or something.

README.md
jeremiah: "using the Yocto project" is better. I'll change.

jeremiah: Done.

README.md
jeremiah: Done

README.md
jeremiah: Done

PR #68, opened 2017-09-05
Changes made after feedback

jeremiah:


jeremiah:
@tobsan How is it possible to update the PR with new commits? I edited the file but on GitHub but the edits didn't show up in the PR. I then rebased my fork and pushed to GitHub but they still didn't show up. There seems to be no obvious way to refresh the PR. Am I missing something?

DunderRoffe:
@jeremiah It should be enough to push the changes to where the PR is syncing against. In this case your master branch. The PR has no content off its own and is only ever the diff between two branches.

So when you did not see any update either you needed to refresh your browser, the push did not really happen or you pushed to the wrong place. My guess is the first option.

Side note: When creating a PR in github you by default give the administrators of the upstream repo push access to the branch you create the PR from. Therefore you might want to consider to create a branch when for the PR.

breakreturn:
Maybe the change was done through the GitHub web interface?

breakreturn:
Err no then there wouldn't have been a PR in the first place I guess :|

jeremiah:
@DunderRoffe I made the changes against my master branch, which was the basis of the PR. In fact I made them in the browser editing directly via GitHub Web UI, but no new syncing of the content occurred. I think in fact that the content is cached and that the diff becomes a thing unto itself. Indeed, looking at the closed PR: https://github.com/Pelagicore/pelux-manifests/pull/65 it shows that the original text is still there. 

You're right I should use a separate branch, I'll try and make sure to do that in the future.

breakreturn:
@jeremiah We usually try to follow this workflow: https://pelagicore.github.io/software-factory/chapters/workflow/git-workflow.html

build-pelux.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
breakreturn: I think we should have the instructions on how to build and use PELUX Baseline in the SWF. I think we should avoid having multiple locations due to the increased maintenance burden. https://pelagicore.github.io/software-factory/chapters/baseplatform/index.html

build-pelux.sh
```diff
@@ -0,0 +1,63 @@
...
+
+sd-image-create ()
+{
+    # hic sunt dracones!
+
+    # Create a new sd image based on the name of the image we're building
+    dd if=/dev/zero of="pelux-intel.sdimg" bs=1024 count=1M
+    sudo losetup -d /dev/loop0
+    sudo losetup /dev/loop0 pelux-intel.sdimg
+    sudo bash ~/pelux/sources/poky/scripts/contrib/mkefidisk.sh /dev/loop0 core-image-pelux-intel-corei7-64.hddimg /dev/mmcblk2
```
jonte: This device path is very hardware specific - not sure we want to hard code such specifics w/o giving an obvious way of tweaking them

build-pelux.sh
```diff
@@ -0,0 +1,63 @@
...
+    sudo losetup -d /dev/loop0
+
+    #
+    DEVICE="/dev/sdb" # get this via command line?
```
jonte: This will wipe /dev/sdb for the user if we're not careful.

build-pelux.sh
jonte: Can't we combine this step with the dd step below? i.e have the mkefidisk.sh script write directly to the hardware device rather than the loopback.

jonte: Also, we should probably rely on the hashbang in mkefidisk.sh rather than assume that bash should be invoked on the script (this is unlikely to cause issues, but technically the script hashbang could change, and a different interpreter could be specified).

jonte: Sorry, the device path being /dev/mmcblk2. I assume this comes from the NUC.

jeremiah: Yeah. I'm sorry that this got included in the PR, it was meant only to be local to my repo since it is a quick and very dirty hack.

jeremiah: Yep, good points.

jeremiah:
The script was a way for me to capture work done, to be an aide-mémoire for the next time I had to do this. I've removed it.

jeremiah:
@jonte @breakreturn PR has been updated following the process outlined and I removed the build script since it wasn't ready for prime time. :-)

breakreturn:
@jeremiah I'll merge this when the CI approves. The first report of a failed job was not correct, it was reported by the wrong CI infra :)

PR #70, opened 2017-09-12
Don't archive binaries for PELUX

tobsan:


Jenkinsfile
```diff
@@ -48,22 +48,24 @@ def buildManifest = {String manifest, String bitbake_image ->
...
-        if (env.ARCHIVE_CACHE) {
-            sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/downloads/ /vagrant/archive/\""
-            sh "vagrant ssh -c \"cp -a ${yoctoDir}/build/sstate-cache/ /vagrant/archive/\""
-        }
-    }
-
-    stage("Archive ${bitbake_image}") {
-        archiveArtifacts artifacts: "archive/**/*"
-    }
+// Use this again when we can give out complete and corresponding source code as well
```
breakreturn: Keeping this for future reference? I don't mind much but I usually get a slap in the face when doing this ;)

DunderRoffe: I say remove it. When we want it back we can just revert the commit.
That is one of the use-cases with having atomic commits :)

breakreturn: @DunderRoffe -<---@

tobsan: Good point, I'll remove it completely instead.

PR #71, opened 2017-09-12
INCOMPLETE: add manifest for rPi with qtauto

tobsan:
Also updates the Jenkinsfile to build rPi version of PELUX


tobsan:
This has been largely outdated due to the pyro branch etc etc. Will create a new one based on the updated layers.

PR #72, opened 2017-09-18
Revert "Jenkinsfile: remove archiving of artifacts."

tobsan:
This reverts commit c0165129f68d86444ffe66865e70f025cff0e36a.

We still save the cache, but we don't publish any artifacts.


breakreturn:
@tobsan So this is a middle step of first reverting before changing so only the cache is saved?

tobsan:
@breakreturn dunno what you mean with middle step. This saves the cache, but does not publish any artifacts.

breakreturn:
Hmm, I'm misreading it then but to me it looks like we're always copying the images and sometimes also the download and sstate caches.

tobsan:
Ah yes, my bad. They are copied but never archived. I'll update so that they're not even copied.

PR #73, opened 2017-10-04
Initial release of pyro based build.

tobsan:
This updates master to point to latest pyro build. This can then be pushed as the pyro branch (for some reason I can't do a PR to create a new branch).

Note that the BSP layers are now included in meta-pelux as of this change.


tobsan:
Also, please don't merge until the CI says it is fine :)

PR #74, opened 2017-10-10
pelux.xml: One manifest for all revisions.

### LINKED ISSUE #24: Refactor descriptions of specific manifests
As of pull-request #23 the description of hardware specific manifests are listed in a way that does not scale very well. Hence, this documentation needs to be refactored.

One suggestion is to use a table witch maps manifests to desired image and hardware target.
Following is an example with the two hardware targets `Foo` and `Bar`:
```
 Hardware | core-image-pelux | core-image-pelux-qt
----------+------------------+--------------------
 Foo      | pelux-foo.xml    | pelux-foo-qt.xml
----------+------------------+--------------------
 Bar      | pelux-bar.xml    | pelux-bar-qt.xml
```

tobsan:
Since we no longer download BSP layers separately (they are part of
meta-pelux), it makes little sense to keep several different variants of
the manifests. The different variants can be chosen after downloading
the layers anyway.

Adds two symlinks for default.xml and manifest.xml that points to
pelux.xml, since default.xml is the default in repo.

Closes #24 


pelux.xml
```diff
@@ -42,16 +42,32 @@
            name="GENIVI/meta-ivi"
            path="sources/meta-ivi" />
 
-  <!-- Qt Support stuff -->
+  <!-- Qt + Qt Automotive support stuff -->
```
GordanM: meta-qt5 is not Qt Auto specific.

tobsan: I know that, but anything below that comment is only needed when we build the QtAS spin, hence the phrasing.

PR #79, opened 2017-10-13
pelux.xml: bump revision for meta-pelux

tobsan:
Will make sure correct image names are shown, among other things.


tobsan:
@DunderRoffe updated commit message.

PR #81, opened 2017-10-23
Vagrant: Remove VirtualBox support

### LINKED ISSUE #64: Vagrant only works with selected versions of VirtualBox
When trying to use the Vagrant based build version of PELUX I failed to get vagrant to run with Docker so I used VirtualBox instead. When running;

$ vagrant up 

I get; 

"The provider 'virtualbox' that was requested to back the machine
'default' is reporting that it isn't usable on this system. The
reason is shown below:

Vagrant has detected that you have a version of VirtualBox installed
that is not supported. Please install one of the supported versions
listed below to use Vagrant:

4.0, 4.1, 4.2, 4.3"

I'm using VirtualBox version 5.1.26r117224 on Debian buster/sid. Do I have to downgrade my VirtualBox to use Vagrant?

tobsan:
We prefer using Docker for builds (faster + lighter) and haven't used
VBox for a while in the CI.

Closes #64 

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,11 +28,14 @@ NeptuneUI and QtApplicationManager.
 
 ### Using vagrant
 
+Please note that we only run this setup in a GNU/Linux system at Pelagicore. It
+should still work under Windows or OSX, we haven't tried it.
+
 Dependencies:
 
 * Vagrant
-* Docker or VirtualBox
-* virtualization enabled in bios
+* Docker
+* Virtualization enabled in BIOS
 
 Procedure:
 
@@ -40,7 +43,7 @@ Procedure:
 2. Start vagrant
 
     ```bash
-    vagrant up --provider="docker"
+    vagrant up
     ```
 
 3. Set variables to be used below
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -28,19 +28,22 @@ NeptuneUI and QtApplicationManager.
...
+* Virtualization enabled in BIOS
 
 Procedure:
 
 1. Clone the pelux-manifests git repository.
 2. Start vagrant
 
     ```bash
-    vagrant up --provider="docker"
+    vagrant up
```
DunderRoffe: Did you verify that not specifying provider works? I have had problem with it

rpannek: I never used the --provider flag and it always worked for me, when did you have problems with that?

Gagi2k: isn't the default provider specified from the base-box ? what's the base-box using ?

tobsan: Yes, I tried it without specifying provider.

DunderRoffe: @rpannek it was some time during the summer I think

PR #82, opened 2017-10-25
Jenkinsfile: build SDK for the images

tobsan:


tobsan:
And the build fails because pa_check fails. Will have to add some pulseaudio libs if we want to build qemu as part of the SDK I think.

PR #84, opened 2017-10-27
Jenkinsfile: some general cleanup

tobsan:
Move the DockerCI node selection into the function that runs the
builds, since it is the same for all builds. Also wrap vagrant calls in
a function for simplicity, now there is no need to escape quotes
everywhere.

Also just playing around with it...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,93 +2,97 @@
 
 // Copyright (C) Pelagicore AB 2017
 
+// Helper function to run commands through vagrant
+def vagrant = {String command ->
+    sh "vagrant ssh -c \"${command}\""
+}
+
 /*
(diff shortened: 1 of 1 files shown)


tobsan:
I just realized that this should be expected to fail until #85 is merged and this is rebased on top because of the issues that #85 solves wrt variant usage in meta-pelux.

From looking at the log, however, the failures has nothing to do with what this PR actually changes in the Jenkinsfile.

tobsan:
Noticed that to build qtwebengine for the pi the multilib libs are needed, apparently. Wait with merge until that change in [vagrant-cookbook/55](https://github.com/Pelagicore/vagrant-cookbook/pull/55/files) is merged.

PR #52, opened 2017-06-20
meta-qt5: update to 5.9.1

GordanM:
Signed-off-by: Gordan Markuš <gordan.markus@pelagicore.com>

PR #89, opened 2017-11-15
Move contents to the SWF, link to it

tobsan:
To avoid double maintenance, all build instructions in the README are
moved to the SWF, and we add a link to the SWF instead. The same is done
with the getting-started.md file.

This is depending on [software-factory #52](https://github.com/Pelagicore/software-factory/pull/52)


README.md
```diff
@@ -5,140 +5,14 @@ hardware targets using the Yocto project.
...
-The script will create configs if there are no configs present, a message about
-created `conf/local.conf` and `conf/bblayers.conf` files is normal.
-
-Finally, build the desired image. See the variables description above for
-information on the different images.
-```bash
-bitbake <image>
-```
+Full documentation of PELUX is available at [the PELUX Software
+Factory](http://pelux.io/software-factory/) Please refer to the Software Factory
```
rpannek: I think I'd prefere a deeper link to at least http://pelux.io/software-factory/chapters/baseplatform/ or even the whole way to http://pelux.io/software-factory/chapters/baseplatform/building-PELUX-sources.html

README.md
```diff
@@ -148,13 +22,14 @@ BSP](https://www.yoctoproject.org/downloads/bsps/pyro23/intel-corei7-64)
 * Minnowboard Max, Turbot
 
 ### PELUX Raspberry Pi 3
-Reference instance for Raspberry Pi 3 (coming soon)
+Reference instance for Raspberry Pi 3. Currently targets 32-bit mode.
```
DunderRoffe: This is a change beyond what the commit message says. Although I agree this is nit picking it is still bad habit!

README.md
tobsan: While that could be useful, I thought not to, since the path might change over time and we would end up with a broken link then.

README.md
tobsan: I'll separate that into another commit.

tobsan: Updated now.

tobsan:
@rpannek @DunderRoffe please review :)

README.md
rpannek: So how does someone understand that they need to click on "Baseplatform" to find anything about bitbake?

tobsan: The frontpage contains a table of contents that has the heading "Building the PELUX sources" in it. I think that suffices.

DunderRoffe:
👍

PR #95, opened 2017-12-15
pelux.xml: bump meta-pelux

tobsan:


tobsan:
@GordanM I suspected something like that might happen. Bumped the cookbook now so hopefully the build goes through.

PR #96, opened 2017-12-19

thenor:

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ from the master branch.
 All feature growth should happen first on the master branch, but will also be
 cherry picked back to the latest yocto release branch. Security and bug fixes
 will be evaluated case by case and backported as necessary. The ambition is to
-actively maintain the four latest releases and/or two year old releases in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,7 +33,7 @@ from the master branch.
 All feature growth should happen first on the master branch, but will also be
 cherry picked back to the latest yocto release branch. Security and bug fixes
 will be evaluated case by case and backported as necessary. The ambition is to
-actively maintain the four latest releases and/or two year old releases in
+actively maintain the two latest releases and/or two year old releases in
```
GordanM: > ... actively maintain the two latest releases and/or two year old releases in ...

I think here it should be `one year old releases` as the Yocto/OE release cadence is 2 releases per year (https://wiki.yoctoproject.org/wiki/Releases).

thenor: Alright. Fixing it.

thenor:
The build was stopped since it is not needed to test a change in the README.

PR #97, opened 2017-12-19

thenor:


thenor:
The build was stopped since it is not needed to test a change in the README.

PR #98, opened 2017-12-19

thenor:


tobsan:
Was too quick perhaps. I don't think we need a PR to put this on pyro. When we release for pyro we should just push master to pyro IMHO. Care to comment @GordanM

thenor:
Pyro will instead be updated with this change by merging master into the branch.

PR #100, opened 2017-12-20
Bump internal layers

thenor:


thenor:
Build stopped on purpose since the change is only a bump of the README files. I will keep a lookout for errors in the nightly.

PR #103, opened 2018-01-02
pelux.xml: bump meta-pelux

tobsan:
Changes in meta-pelux from earlier revisions are:
* systemd-networkd: add default DHCP for ethernet

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -23,7 +23,7 @@
 
   <project remote="github"
            revision="28f52b1451c6bebe46224b81dc6849babf286ce1"
-           upstream="master"
+           upstream="pyro"
            name="Pelagicore/meta-bistro"
            path="sources/meta-bistro" />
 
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -33,8 +33,8 @@
            path="sources/meta-virtualization"/>
 
   <project remote="github"
-           revision="e87ed59b1cae0939a9874f1089941f85c0992292"
-           upstream="master"
+           revision="40d0c53b0e0adae35712fac5396572d2fd28dbf5"
+           upstream="pyro"
```
GordanM: So this is a commit to the manifests master right? Shouldn't that manifest version always follow master meta-bistro and meta-pelux? Because eventually it will become rocko and so on?

tobsan: I gave this some thought. Given that master and pyro are aligning currently, I think we can follow the same pattern as we do on meta-bistro and meta-pelux. This way, we can push master to pyro in this manifest-repo as well and it will just work. After release of pyro we can bump to rocko here (which will then follow master in the layers).

On a larger scale of things, we talked earlier about always following the latest on a branch for the repos we control ourselves, but I think we can wait with that until rocko (so after PELUX 1.0)

GordanM: I see, valid point. I guess it has come to that point that we will be releasing so it makes sense to make master closer to pyro.

PR #105, opened 2018-01-10
Extend docker from crops yocto image

chuanjin:
Signed-off-by: Chuan Jin <cjin@luxoft.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,13 @@
-FROM ubuntu:16.04
+FROM crops/yocto:ubuntu-16.04-base
 
-ARG userid=1000
+USER root
 
 RUN mkdir /var/run/sshd
 
 # Install dependencies in one command to avoid potential use of previous cache
 # like explained here: https://stackoverflow.com/a/37727984
 RUN apt-get update \
-    && apt-get install -y openssh-server sudo cpio inetutils-ping locales
+    && apt-get install -y openssh-server inetutils-ping cvs subversion python-pysqlite2 help2man libxml2-utils libsdl1.2-dev graphviz qemu-user g++-multilib curl
     
 RUN rm -rf /var/lib/apt/lists/*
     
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -7,18 +9,21 @@ RUN mkdir /var/run/sshd
...
     
 RUN rm -rf /var/lib/apt/lists/*
     
 # en_US.utf8 is required by Yocto sanity check
 RUN localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 ENV LANG en_US.utf8
 
+RUN userdel -r yoctouser
 RUN useradd --uid $userid vagrant --create-home --user-group --groups sudo
 RUN echo 'vagrant:vagrant' | chpasswd
```
tobsan: Is it possible to tell vagrant to use a different user than "vagrant"? In that case it would be nicer to be able to still use "yoctouser"

Vagrantfile
```diff
@@ -35,7 +32,7 @@ Vagrant.configure(2) do |config|
     # Configure username and password in git
     config.vm.provision "shell", privileged: false, path: "vagrant-cookbook/system-config/vagrant-ssh-user.sh"
 
-    # Configure username and password in git
+    # Download and set up the repo tool
```
tobsan: This should go into a separate commit imho.

Dockerfile
chuanjin: Yes, then I replace vagrant with yoctouser

chuanjin:
Same build error, maybe disable cache?

tobsan:
This build error is really weird since it seems the Jenkinsfile doesn't pick up on the changed variable content. We tried removing all docker images and their vagrant contexts but still same result. @rpannek do you have any idea what it could be?

rpannek:
Just for the record since we talked about it in between, because @chuanjin is not in the Pelagicore organization he isn't allowed to run a job with a changed Jenkinsfile.

PR #109, opened 2018-01-17
Move more contents from Vagrantfile to Dockerfile

tobsan:
Please do not merge before the CI has run.

As we are transitioning away from using Vagrant, there are now only
settings for synced folders and building the docker image left in the
Vagrantfile. These can probably be moved to be done by Jenkins later on.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,23 @@ USER root
 
 # Install dependencies in one command to avoid potential use of previous cache
 # like explained here: https://stackoverflow.com/a/37727984
-RUN apt-get update \
-    && apt-get install -y openssh-server inetutils-ping iptables cvs subversion coreutils python3-pip libfdt1 python-pysqlite2 help2man libxml2-utils libsdl1.2-dev graphviz qemu-user g++-multilib curl
+RUN apt-get update && apt-get install -y \
+        openssh-server \
+        inetutils-ping \
+        iptables \
+        cvs \
+        subversion \
+        coreutils \
+        python3-pip \
+        libfdt1 \
+        python-pysqlite2 \
+        help2man \
+        libxml2-utils \
+        libsdl1.2-dev \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,28 +1,60 @@
...
 
 USER root
 
-RUN mkdir /var/run/sshd
-
 # Install dependencies in one command to avoid potential use of previous cache
 # like explained here: https://stackoverflow.com/a/37727984
-RUN apt-get update \
-    && apt-get install -y openssh-server inetutils-ping iptables cvs subversion coreutils python3-pip libfdt1 python-pysqlite2 help2man libxml2-utils libsdl1.2-dev graphviz qemu-user g++-multilib curl
+RUN apt-get update && apt-get install -y \
```
GordanM: Unrelated change.

tobsan: True that. Would you rather see it separated?

GordanM: Yeah that would be good, because applying patches (if it comes to that) might prove to be tricky later on.

GordanM: Thanks!

Dockerfile
```diff
@@ -1,28 +1,60 @@
...
+        python3-pip \
+        libfdt1 \
+        python-pysqlite2 \
+        help2man \
+        libxml2-utils \
+        libsdl1.2-dev \
+        graphviz \
+        qemu-user \
+        g++-multilib \
+        curl
```
DunderRoffe: Add a \ at the end of this row as well, then you can mix and match easier

tobsan: Adding a backslash there makes the Dockerfile unparseable, so no.

tobsan: Or rather, it gives a warning that will become an error in future versions.

DunderRoffe: :/

PR #112, opened 2018-01-23
pelux.xml: bump to rocko

tobsan:

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -12,66 +12,73 @@
 
   <!-- Base stuff -->
   <project remote="oe"
-           revision="5e82995148a2844c6f483ae5ddd1438d87ea9fb7"
+           upstream="rocko"
+           revision="6e3fc5b8d904d06e3aa77e9ec9968ab37a798188"
            name="meta-openembedded"
            path="sources/meta-openembedded"/>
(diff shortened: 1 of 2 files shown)


pelux.xml
```diff
@@ -12,66 +12,70 @@
...
-           revision="717303e6fbcbbe181ad9645d762eb5a85d934523"
+           upstream="rocko"
+           revision="a03c452917842b15451ad7098c1216eaec09cd2b"
            name="poky"
            path="sources/poky"/>
 
   <project remote="github"
-           revision="28f52b1451c6bebe46224b81dc6849babf286ce1"
-           upstream="pyro"
+           revision="rocko"
```
thenor: Shouldn't this point to a commit ID?
Same for meta-pelux? Or do you want them to not be connected to a commit for now?

tobsan: This is stated in the commit message :)

tobsan:
Let's wait for the CI to finish. It already caught one major issue so I would like to be sure.

tobsan:
After discussions on IRC, updated so that we don't follow latest in our own layers, since we don't have the CI set up to handle that properly.

PR #114, opened 2018-01-24
WIP: pelux.xml: bump meta-pelux

thenor:
Adds the following commits:
* packagegroup: Change to use debug-utils for dev images

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,89 +2,99 @@
 
 // Copyright (C) Pelagicore AB 2017
 
-/*
- * For supported variant names, see conf/variants/ in meta-pelux
- */
-void buildManifest(String variant_name, String bitbake_image) {
+// Helper function to run commands through vagrant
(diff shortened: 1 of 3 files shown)


thenor:
Awaiting the bigger bump in master before starting the review.

pelux.xml
```diff
@@ -23,7 +23,7 @@
 
   <project remote="github"
            revision="28f52b1451c6bebe46224b81dc6849babf286ce1"
-           upstream="pyro"
+           upstream="master"
```
tobsan: Why this change when bumping meta-pelux?

thenor: I didn't realize until after I had made the PR that another PR bumped the entire internal layers, and also updated the "upstream" branch name. It is a typo on master.

thenor:
I accidentally closed this, but I will need to rebase when the bump to rocko is done, so I will make a new PR when that is done.

PR #117, opened 2018-01-25
WIP: Jenkinsfile: sync CACHE over SSH instead of NFS

tobsan:
We still read the cache over NFS, but with SSH to write we don't need
uids to match, which means more possible freedom in for example the
Dockerfile.


tobsan:
TODO: How to get SSH keys in a nice way into the container? Or perhaps we can actually use Vagrant for something here?

PR #119, opened 2018-01-28
pelux.xml: Add meta-swupdate

FlorentRevest:
In order to integrate swupdate to PELUX, this PR downloads meta-pelux into the sources/ directory.

This PR is written for pyro. If https://github.com/Pelagicore/pelux-manifests/pull/112 is merged before, I will update the current commit hash and branch label to point to the rocko branch of meta-swupdate.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -41,6 +41,12 @@
            name="meta-virtualization"
            path="sources/meta-virtualization"/>
 
+  <project remote="github"
+           revision="00ff53a574bf03bed29d80945f5e589879d3e2b2"
+           upstream="rocko"
+           name="sbabic/meta-swupdate"
+           path="sources/meta-swupdate" />
+
   <!-- GENIVI stuff, update the upstream from master to 14.0 when possible-->
   <project remote="github"
            upstream="master"


GordanM:
AFAIK, new features should be developed against rocko (master) but I guess @tobsan and @thenor could give you more input on that, I might be mistaken. Unfortunately, there are some problems with the docker containers or the slaves in our CI system so the tests are failing.

thenor:
@FlorentRevest We have been having issues on the pelux-manifests CI and that is the reason why the rocko update is not on master yet. Tobias reverted a change that seemed to have introduced the issue, but it didn't solve it completely. I guess nothing is stopping other commits until the "bump to rocko" is merged, but just so you know the builds can fail for seemingly no reason.

tobsan:
@FlorentRevest can you rebase this on top of master and target rocko instead, now that it has been merged? Thanks!

tobsan:
Ehm...by target I mean yocto target, not branch! We still want it on master, but master currently targets rocko :D

FlorentRevest:
tobsan: yep, I did it tonight and compiled while I was sleeping. I'm gonna check the results and send a PR once I'll be back home !

FlorentRevest:
There you go!
https://github.com/Pelagicore/meta-pelux/pull/89 still applies cleanly and has been tested manually to not break anything.
https://github.com/Pelagicore/pelux.io/pull/61 can then be merged as well.

We can wait for the CI to finish but since it probably already passed the repo commands, this should be ok.

PR #123, opened 2018-02-12
WIP: pelux.xml: bump meta-pelux

thenor:
Adds the following commits for meta-pelux:
    * variant/common: Don't install extra devel deps for QEMU
    * packagegroup: Change to use debug-utils for dev images
    
Adds the following commits for meta-bistro:
    * Remove bitbake patch, upstreamed into rocko
    * packagegroups: Split into debug utils and utils

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -24,28 +24,28 @@
            path="sources/poky"/>
 
   <project remote="github"
-           upstream="rocko"
-           revision="7404235400b8e4de2d85c07ec5d4c6b8ac56e97d"
+           upstream="master"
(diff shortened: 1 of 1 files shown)


thenor:
I don't like the idea of having to back out meta-swupdate, since it seems to have worked previously.

thenor:
The build failures started when I included the meta-swupdate commits. I then tried to only change the meta-pelux revision, but still had trouble. When I removed the meta-swupdate from the manifest I got a successful build. I will try again now that updates have been made in meta-pelux. Hopefully the build will not fail again.

PR #124, opened 2018-02-27
Archive artifacts and test the SDK

tobsan:


sdk/install_sdk
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
demonicblue: Place a copyright notice here

tobsan: Done!

PR #125, opened 2018-02-28
Jenkinsfile: always archive cache

tobsan:
We want to store cache even if the build failed in the middle, because
everything up to the failure went well, meaning those downloads and
sstate-cached binaries are fine.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -68,22 +68,26 @@ void buildManifest(String variant_name, String bitbake_image) {
             }
         }
 
-        stage("Bitbake ${variant_name}") {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -68,28 +68,34 @@ void buildManifest(String variant_name, String bitbake_image) {
...
+                }
             }
         }
 
         // Always try to shut down the machine
         // Shutdown the machine
-        sh "vagrant destroy -f || true"
+        retry(5) {
+            sh "vagrant destroy -f || true"
+        }
```
thenor: How long does it take before the retries? Is it in quick succession?

tobsan: It would retry directly if the step fails, but since we or with true I'm not really sure what I was thinking. I'll revert this part of the changes.

PR #126, opened 2018-03-13
Smoke testing for Qemu

thenor:
Adds a script for smoke testing triggered from the Jenkinsfile.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -38,6 +38,16 @@ parallel (
             def manifests = load "ci-scripts/yocto.groovy"
             manifests.buildManifest("rpi-qtauto", "core-image-pelux-qtauto-neptune")
         }
+    },
+    'qemu': {
+        node("Yocto") {
+            boolean nightly = env.NIGHTLY_BUILD == "true"
+            boolean smokeTests = env.SMOKE_TEST == "true"
+            if (nightly) {
+                checkout scm
+                def manifests = load "ci-scripts/yocto.groovy"
+                manifests.buildManifest("qemu-x86-64_nogfx", "core-image-pelux-minimal", smokeTests)
+            }
+        }
     }
 )
-
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -104,6 +104,7 @@ void runSmokeTests(String yoctoDir, String imageName) {
         }
     } catch(e) {
         echo "There were failing tests"
+        println(e.getMessage())
     } finally {
         stage("Publish smoke test results") {
             reportsDir="/vagrant/${archiveDir}/test_reports/${imageName}/"
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -76,7 +76,7 @@ void buildImageAndSDK(String yoctoDir, String imageName, boolean dev=true) {
     // means we should do a fetchall.
     if (vagrant("test -f ${yoctoDir}/build/conf/site.conf", true) != 0) {
         stage("Fetch sources") {
-            vagrant("/vagrant/cookbook/yocto/fetch-sources-for-recipes.sh ${yoctoDir} ${bitbake_image}")
+            vagrant("/vagrant/cookbook/yocto/fetch-sources-for-recipes.sh ${yoctoDir} ${imageName}")
         }
     }
 
@@ -138,7 +138,7 @@ void archiveArtifacts(String yoctoDir, String suffix) {
     }
 }
 
-void buildWithLayer(String variant_name, String bitbake_image, String layer, String layerPath) {
+void buildWithLayer(String variantName, String imageName, String layer, String layerPath) {
     // Store the directory we are executed in as our workspace.
     String yoctoDir = "/home/yoctouser/pelux_yocto"
     String manifest = "pelux.xml"
@@ -154,13 +154,13 @@ void buildWithLayer(String variant_name, String bitbake_image, String layer, Str
         replaceLayer(yoctoDir, layer, layerPath)
 
         // Setup yocto
-        String templateConf="${yoctoDir}/sources/meta-pelux/conf/variant/${variant_name}"
+        String templateConf="${yoctoDir}/sources/meta-pelux/conf/variant/${variantName}"
         setupBitbake(yoctoDir, templateConf)
         setupCache(yoctoDir, env.YOCTO_CACHE_URL)
 
         // Build the images
         try {
-            buildImageAndSDK(yoctoDir, bitbake_image)
+            buildImageAndSDK(yoctoDir, imageName)
         } finally { // Archive cache even if there were errors.
             boolean archive = env.ARCHIVE_CACHE == "true"
             archiveCache(yoctoDir, archive, env.YOCTO_CACHE_ARCHIVE_PATH)
@@ -176,7 +176,7 @@ void replaceLayer(String yoctoDir, String layerName, String newPath) {
     vagrant("mv /vagrant/${layerName} ${yoctoDir}/sources/")
 }
 
-void buildManifest(String variant_name, String bitbake_image, boolean smokeTests=false) {
+void buildManifest(String variantName, String imageName, boolean smokeTests=false) {
     // Store the directory we are executed in as our workspace.
     String yoctoDir = "/home/yoctouser/pelux_yocto"
     String manifest = "pelux.xml"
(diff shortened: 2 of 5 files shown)


Jenkinsfile
```diff
@@ -68,6 +68,13 @@ void buildManifest(String variant_name, String bitbake_image) {
             }
         }
 
+        if ("smoke-test" in options || "export-oe-tests" in options) {
```
erikboto: In another project we use a Set called "options" which is passed as a parameter to buildManifest. That Set it not available here and that's what the build is failing on.

Jenkinsfile
```diff
@@ -91,6 +98,75 @@ void buildManifest(String variant_name, String bitbake_image) {
             }
         }
 
+        if ("smoke-test" in options) {
```
erikboto: Same issue as line 71.

Jenkinsfile
thenor: Thanks! I realized that this is why is fails, but was trying to find "options". Do you have some time for help me, @erikboto? @rpannek started to help me but is not in the office now.

Jenkinsfile
```diff
@@ -91,6 +98,77 @@ void buildManifest(String variant_name, String bitbake_image) {
             }
         }
 
+        if (variant_name == "qemu-x86-64_nogfx") {
+            try {
```
tobsan: Perhaps all this could be a separate Jenkinsfile that we load and run if needed? Since it is such a different mode of operation I think that would make sense.

thenor: That sounds like a good idea. For now this is only a WIP, but I will break it out to finish the user story.

thenor:
@erikboto Maybe I can give you access to my repo so you can push your changes to this branch and PR?

test-scripts/local.conf.appendix
```diff
@@ -0,0 +1,12 @@
...
+    testimage \
+    testexport \
+"
+
+# All test suites to run
+TEST_SUITES = " \
+    ping \
+    ssh \
+    auto \
+"
```
erikboto: This file can be removed right, there's another version with the same name that I think is the one used.

local.conf.appendix
```diff
@@ -1,3 +1,20 @@
+# Enable testing
 
+# Needed since a bbappend in meta-ivi adds libusb PACKAGECONFIG which is 
+# broken for some weird reason
+BBMASK .= "|meta-ivi/recipes-devtools/qemu/"
```
erikboto: Maybe this should be done not only when testing? So moved to the meta-pelux/variant/common/local.conf?

test-scripts/local.conf.appendix
thenor: Oops. A merge mistake.

local.conf.appendix
thenor: @tobsan what do you think?

tobsan: If it is a generic setting that we always want to have enabled, then it should of course go in ``meta-pelux/variant/common/local.conf``

GordanM: Furthermore, shouldn't there be a change upstream to meta-ivi that fixes this `PACKAGECONFIG`? I guess it would be nice to inform them about this. Does the downstream project experience these side effects as well?

thenor: @GordanM You were removing a bunch of qemu stuff. This is still ok to include in our general config?

thenor: I guess there are two topics being discussed now. The upstreaming and solving of the qemu devtool from meta-ivi and if we should include this change in the local.conf now.

GordanM: @thenor I've removed some meta-bistro QEMU startup scripts that are not being used anywhere, should not impact any functionality. 

Nice! That is good to know. We can always create two separate changes for this, maybe this even belongs in the temp patches then? You know best, I've just wanted to point this out :).

local.conf.appendix
```diff
@@ -1,3 +1,20 @@
...
+
```
thenor: This and the rest of the added lines I will keep only in the test-scripts directory.

test-scripts/local.conf.appendix
thenor: My idea here was to have one file for the test specific env. Maybe it's overkill? What do you think @tobsan (or others)?

thenor:
Is there a way to make this build as if it was the nightly? @tobias, do you know? Or should I wait for the next nightly after it is merged to see that it passes?

thenor:
I got the connection refused issue on the rpi build, but all other images look fine.

GordanM:
I believe this is good enough right? Because there is no impact on the rpi image :)

ci-scripts/yocto.groovy
```diff
@@ -87,6 +95,25 @@ void buildImageAndSDK(String yoctoDir, String imageName, boolean dev=true) {
...
 }
 
+void runSmokeTests(String yoctoDir, String imageName) {
+    String archiveDir = "testReports-" + imageName
+
+    try {
+        stage("Perform smoke testing") {
+            vagrant("/vagrant/cookbook/yocto/runqemu-smoke-test.sh ${yoctoDir} ${imageName}")
+        }
+    } catch(e) {
```
rpannek: I really dislike this way of catching everything and then not even printing what actually happened because this makes impossible to find what actually happened. Either remove the catch-all or make it more specific for some type of a error and print the call stack or dive some possibility to find out what happened instead of just printing a string.

local.conf.appendix
```diff
@@ -1,3 +1,6 @@
+# Needed since a bbappend in meta-ivi adds libusb PACKAGECONFIG which is
+# broken for some weird reason
+BBMASK .= "|meta-ivi/recipes-devtools/qemu/"
```
rpannek: Did we report a bug?

test-scripts/local.conf.appendix
rpannek: A more specific one is nice but I guess only one should be used?

local.conf.appendix
thenor: Not yet. I will discuss this with @frznlogic and see if we can find a way forward (either we find the solution or we just report it as a bug).

test-scripts/local.conf.appendix
thenor: There is configuration needed for the tests, but not for anything else. That was shy it felt most reasonable for me to have a new conf file for the tests.

ci-scripts/yocto.groovy
thenor: See the newly added code below, @rpannek. Is this what you had in mind?

local.conf.appendix
thenor: I have talked to @frznlogic and we will look into this. It will not be done within this PR though.

rpannek: Sure, just so we report the bug at least.

thenor:
@rpannek @GordanM What do you think about keeping the Qemu flag that I introduced in the last commit. Should I keep it or remove it?

Jenkinsfile
```diff
@@ -10,34 +10,56 @@
 parallel (
     'intel': {
         node("Yocto") {
-            checkout scm
-            def manifests = load "ci-scripts/yocto.groovy"
-            manifests.buildManifest("intel", "core-image-pelux-minimal")
+            boolean only_qemu = env.QEMU == "true"
```
rpannek: That is a really weird way of implementing that by copy&pasting it that many times.

I would not try to do a special thing for qemu, but instead abstract it away so it is easy to extend for example you could create this map programatically before the parallel() call. And there you could have a list of the jobs which are default (which would be all of them) and then prepare so one can manipulate this list to remove some of them, or overwrite it with only qemu in it on demand.

This way you would make it possible to just run the rpi build ord only the qemu and intel-qtauto, or any other combination without every time adding special if clauses.

local.conf.appendix
frznlogic: sorry, I missed this discussion earlier. I did look at it briefly, the PACKAGECONFIG is super old... I'll remove it and see what people/the build system think of it.

Jenkinsfile
thenor: I guess you mean this as an improvement to the Jenkinsfile, so not actually within the scope of adding smoke tests. It does make sense to me to try to get that to work.

thenor: I can remove the qemu flag. I just wanted to use it for being able to run from Jenkins in a simple way without hard-coding.

thenor: Sorry. I misunderstood your comment. Yes, I agree this is not a nice way of doing this.

thenor:
@rpannek I removed the flag I had introduced and we will introduce a nicer way to do this using maybe a list sent from Jenkins, but not included in the scope of this pull request.

PR #130, opened 2018-03-28
pelux.xml: bump most layers

tobsan:
I've test built locally and it all passes, so I hope Jenkins agree with me on that. See individual commits for details.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -2,7 +2,7 @@
 <manifest>
 
   <!-- Settings -->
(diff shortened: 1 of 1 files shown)


tobsan:
This fails due to a bump in bistro. There is a fix in bistro on https://github.com/Pelagicore/meta-bistro/pull/158

GordanM:
@tobsan I've merge the bistro commit. Sorry about that.

PR #131, opened 2018-04-03
Jenkinsfile: more measures to ensure cleanup

tobsan:
The whole build is put in a try/finally step, and even if the archiving
of cache/artifacts fail we still shutdown the vagrant machine in an
outer finally block.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,87 +28,90 @@ void buildManifest(String variant_name, String bitbake_image) {
         // Stages are subtasks that will be shown as subsections of the finished
         // build in Jenkins.
 
-        stage("Checkout ${variant_name}") {
-            // Checkout the git repository and refspec pointed to by jenkins
-            checkout scm
-            // Update the submodules in the repository.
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -28,87 +28,90 @@ void buildManifest(String variant_name, String bitbake_image) {
...
+                        vagrant("/vagrant/ci-scripts/copy_to_archive ${yoctoDir}/build /vagrant/${artifactDir}")
 
-                // And save them in Jenkins
-                archiveArtifacts "${artifactDir}/**"
+                        // And save them in Jenkins
+                        archiveArtifacts "${artifactDir}/**"
+                    }
+                }
+            }
+        } finally { // Archive cache even if there were errors.
```
rpannek: The code down there doesn't archive anything it just destroys the vagrand machine, I'm not sure this comment makes sense here.

tobsan: No, the comment is out of place. Will update!

PR #132, opened 2018-04-05
Jenkinsfile: promote environment for NIGHTLY build

tobsan:
Apparently, the environment directive removes all other env variables.


tobsan:
Perhaps it would be better to just remove the whole ``environment { }`` block.

PR #133, opened 2018-04-09
Break out groovy code from Jenkinsfile to facilitate reuse

tobsan:
As seen, the Jenkinsfile is now a "load" directive and calling a function on the loaded file object. A similar pattern will be used to build pelux-manifests from pull requests on meta-pelux, meta-bistro and meta-template (but these will have to clone this repo first).

Example (unfinished) for meta-pelux: https://github.com/Pelagicore/meta-pelux/pull/106/files

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,31 +2,42 @@
 
 // Copyright (C) Luxoft Sweden AB 2018
 
-def manifests = load "ci-scripts/yocto.groovy"
-
 // Run the different variants in parallel, on different slaves (if possible)
+
+// We could do a single "checkout scm" and load the groovy script, but to do
+// that one has to allocate a node first, which would stay busy until the end of
+// the parallel pipeline
 parallel (
     'intel': {
         node("Yocto") {
+            checkout scm
+            def manifests = load "ci-scripts/yocto.groovy"
             manifests.buildManifest("intel", "core-image-pelux-minimal")
         }
     },
 
     'intel-qtauto': {
         node("Yocto") {
+            checkout scm
+            def manifests = load "ci-scripts/yocto.groovy"
             manifests.buildManifest("intel-qtauto", "core-image-pelux-qtauto-neptune")
         }
     },
 
     'rpi': {
         node("Yocto") {
+            checkout scm
+            def manifests = load "ci-scripts/yocto.groovy"
             manifests.buildManifest("rpi", "core-image-pelux-minimal")
         }
     },
 
     'rpi-qtauto': {
         node("Yocto") {
+            checkout scm
+            def manifests = load "ci-scripts/yocto.groovy"
             manifests.buildManifest("rpi-qtauto", "core-image-pelux-qtauto-neptune")
         }
     }
 )
+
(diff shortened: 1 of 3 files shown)


ci-scripts/yocto.groovy
```diff
@@ -0,0 +1,189 @@
...
+        // Setup yocto
+        String templateConf="${yoctoDir}/sources/meta-pelux/conf/variant/${variant_name}"
+        setupBitbake(yoctoDir, templateConf)
+        setupCache(yoctoDir, env.YOCTO_CACHE_URL)
+
+        // Build the images
+        try {
+            buildImageAndSDK(yoctoDir, bitbake_image)
+        } finally { // Archive cache even if there were errors.
+            boolean archive = env.ARCHIVE_CACHE == "true"
```
thenor: In another change you rewrote this as "env.VAR?.trim" shouldn't that be done here?
(And also for other env vars further down in the code)

Jenkinsfile
```diff
@@ -1,124 +1,36 @@
 #!/usr/bin/groovy
 
-// Copyright (C) Pelagicore AB 2017
+// Copyright (C) Luxoft Sweden AB 2018
```
thenor: :+1:

ci-scripts/yocto.groovy
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/groovy
+
+// Copyright (C) Pelagicore AB 2017
```
thenor: Luxoft 2018

ci-scripts/yocto.groovy
tobsan: It could have been done here as well, yes. I felt however that being explicit about stuff being booleans adds some certainty.

tobsan:
@GordanM fixed!

ci-scripts/yocto.groovy
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+void replaceLayer(String yoctoDir, String layerName, String newPath) {
+    vagrant("rm -rf ${yoctoDir}/sources/${layerName}")
+    sh "cp -R ${newPath} ${layerName}"
+    vagrant("mv /vagrant/${layerName} ${yoctoDir}/sources/")
+}
+
+void buildManifest(String variant_name, String bitbake_image) {
```
rpannek: Wouldn't it make sense to only have one function which would do both instead of copy&pasting most of the code?

Jenkinsfile
```diff
@@ -1,117 +1,43 @@
...
```
frznlogic: Actually, is this correct? Shouldn't the old copyright still be there?

ci-scripts/yocto.groovy
tobsan: I had that at first, but the code became too cluttered and hard to read. And since most of its steps (except the parts that differ) are just calls to the same functions, I figured this was fine.

Jenkinsfile
tobsan: This is technically correct since Pelagicore AB changed name to Luxoft Sweden AB. And while changing the name I updated the year since we are now in 2018.

ci-scripts/yocto.groovy
rpannek: Hm ok I see. Ok.

PR #135, opened 2018-04-24
yocto.groovy: Build SWUpdate artifacts for the RaspberryPi 3

FlorentRevest:
This introduces CI support for the core-image-pelux-*-update recipes.
Those recipes are currently only available for the Raspberry Pi 3
therefore a check is done to only build when variant_name starts with
rpi.

Signed-off-by: Florent Revest <revestflo@gmail.com>

/!\ This patch depends on https://github.com/Pelagicore/meta-pelux/pull/113


ci-scripts/yocto.groovy
```diff
@@ -77,6 +77,9 @@ void buildImageAndSDK(String yoctoDir, String imageName, boolean dev=true) {
         if (dev) {
             vagrant("/vagrant/cookbook/yocto/build-images.sh ${yoctoDir} ${imageName}-dev")
         }
+        if (update) {
+            vagrant("/vagrant/cookbook/yocto/build-images.sh ${yoctoDir} ${imageName}-update")
+        }
```
frznlogic: Actually, this is a bit weird imho... what happens if you want to build an update to a -dev image? Is that possible?

FlorentRevest: Correct, actually this is possible https://github.com/FlorentRevest/meta-pelux/tree/generic-swupdate-artifacts/recipes-extended/images
I could add an if (dev) then build the image-dev-update artifcat as well

frznlogic: I guess you need to take it into account somehow at least :)

FlorentRevest:
Has anyone seen those CI errors before?

TSAnsari:
Haven't seen before. May be @thenor has seen it. It may be related to Docker image related issue.

I see the following error complaining that the container is running when vagrant tries to shutdown.

"Stderr: Error response from daemon: conflict: unable to delete dde9a839c323 (cannot be forced) - image is being used by running container 06cbcaa4847e"

On the other hand it fails initially because the image didn't come up.

thenor:
We seem to be having several Docker issues at the moment. Hard to say if they are related. The Docker "connection refused" (something docker socket) has been possible to solve with the nightly prune Jenkins job. I have even rerun that job two times today and been able to remove some stale containers and images. The "unable to delete"/"image is being used" is also usually solved by rerunning the build. I have no better solution at this moment.

thenor:
If you can see that it builds fine when you don't get the environment issues then I don't see why we can't merge.

PR #136, opened 2018-05-11
Add CI jobs for ARP target

SKumarDas:
Adding CI jobs for targets "arp" and "arp-qtauto" respectively.

Signed-off-by: Sapan Das <SKumarDas@luxoft.com>


Jenkinsfile
```diff
@@ -39,5 +39,21 @@ parallel (
             manifests.buildManifest("rpi-qtauto", "core-image-pelux-qtauto-neptune")
         }
     }
```
thenor: You are missing a comma here.
Is this rebased to the latest? I think there should also be a qemu image here?

PR #137, opened 2018-05-15
ci-scripts/do_repo_init: suppress error message

erikboto:
The error from deleting a branch that doesn't exist is safe, but the
error show up in logs which is annoying when looking for true errors.

### DIFF
--- a/ci-scripts/do_repo_init
+++ b/ci-scripts/do_repo_init
@@ -51,7 +51,7 @@ cp -r ${SYNC_DIR} ${COPY_DIR}
 
 # Make sure there is a master branch and that it is pointing to the latest commit.
 cd ${COPY_DIR}
-git branch -D master || true
+git branch -D master 2> /dev/null || true
 git checkout -b master
 cd -
 


ci-scripts/do_repo_init
```diff
@@ -51,7 +51,7 @@ cp -r ${SYNC_DIR} ${COPY_DIR}
 
 # Make sure there is a master branch and that it is pointing to the latest commit.
 cd ${COPY_DIR}
-git branch -D master || true
+git branch -D master 2> /dev/null || true
 git checkout -b master
```
TSAnsari: The error message also appears when "git checkout -b master" fails due to already existing master branch.  This may also be suppressed. See the error msg below,
"fatal: A branch named 'master' already exists."

erikboto: @TSAnsari What do you mean? How can the branch exist if we delete it on the line before?

erikboto: Also, if there was an error in the "git checkout -b master" row we want that to be visible and causing the build to fail.

erikboto: If that's the case I guess it should be handled too by the script. I guess we could do something like:
git checkout -b superrandombranch
git branch -D master 2> /dev/null
git checkout -b master

TSAnsari: Perhaps a uuid as superrandombranch just to make sure no conflicts happen?

PR #138, opened 2018-05-15
Remove temporary fixes for meta-ivi libusb dependency

thenor:


thenor:
This includes a bump to 14.0 for meta-ivi, so I don't want to merge this unless we decide to do this bump at this time. I want to test the upstream fix for a libusb dependency which is unnecessary.

thenor:
The build failed on the known gammaray issue, so if approved I guess we can merge even though we don't have a 100% successful build.

PR #141, opened 2018-05-17
CI: add missing smoke test flag in yocto.groovy

thenor:
The smokeTests flag did not have a default which caused errors
when using buildWithLayer.


TSAnsari:
Looks good to me

PR #146, opened 2018-05-25
Refactor Jenkinsfile

DunderRoffe:
Extract duplicate code to reusable function

Signed-off-by: Viktor Sjölind <vsjolind@luxoft.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -7,75 +7,58 @@
 // We could do a single "checkout scm" and load the groovy script, but to do
 // that one has to allocate a node first, which would stay busy until the end of
 // the parallel pipeline
+
+void buildOnYoctoNode(String variant, String image, boolean checkSmokeTests = false) {
+    node("Yocto") {
+        boolean smokeTests = false
+
+        if (checkSmokeTests) {
+            smokeTests = env.SMOKE_TEST == "true"
+        }
+
+        checkout scm
+        def manifests = load "ci-scripts/yocto.groovy"
+        manifests.buildManifest(variant, image, smokeTests)
+    }
+}
+
+void buildOnYoctoNodeNightly(String variant, String image, boolean checkSmokeTests = false) {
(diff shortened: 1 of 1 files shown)


DunderRoffe:
I'm not sure this works but lets see what the CI says :)

Jenkinsfile
```diff
@@ -7,75 +7,58 @@
...
-            boolean smokeTests = env.SMOKE_TEST == "true"
-            if (nightly) {
-                checkout scm
-                def manifests = load "ci-scripts/yocto.groovy"
-                manifests.buildManifest("qemu-x86-64_nogfx", "core-image-pelux-minimal", smokeTests)
-            } else {
-                println("Nothing to do for qemu")
-            }
-        }
+        buildOnYoctoNodeNightly("qemu", "core-image-pelux-minimal", true)
```
thenor: This "true" is a bit of a magic value. I would have preferred to show what it represents. Also it used to be set in a environment var in Jenkins.

Jenkinsfile
```diff
@@ -7,75 +7,58 @@
...
+        def manifests = load "ci-scripts/yocto.groovy"
+        manifests.buildManifest(variant, image, smokeTests)
+    }
+}
+
+void buildOnYoctoNodeNightly(String variant, String image, boolean checkSmokeTests = false) {
+    boolean nightlyBuild = env.NIGHTLY_BUILD == "true"
+
+    if (nightlyBuild) {
+        buildOnYoctoNode(variant, image, checkSmokeTests)
```
thenor: The check of this env var has been removed, so the behavior is no longer as it used to be. This should be changed so it checks the env var as the buildOnYoctoNode function does.

thenor: Oh never mind it is actually checked since this calls the other function. However you do set it always when running qemu, which is not how it used to be.

PR #154, opened 2018-06-04
Smoke test doc

thenor:
This is just to have some type of clarification for the conf file in test-scripts. A bit superfluous perhaps.


thenor:
I will document this in a different way.

PR #159, opened 2018-06-12
Bump meta-pelux and meta-arp

TSAnsari:


sashko:
@TSAnsari ping; what's up with this change?

TSAnsari:
This should be discarded. This bump was done in a separate commit by fisnik, so we don't need this one. I will discard it

PR #165, opened 2018-06-20
pelux.xml: bump meta-arp and meta-pelux

fhajredini:


TSAnsari:
Did you tested the build that it builds with these new revisions?

PR #171, opened 2018-07-05
pelux.xml: Bump meta-qt5 and meta-boot2qt to 5.11

fhajredini:

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -56,15 +56,15 @@
 
   <!-- Qt + Qt Automotive support stuff -->
   <project remote="code.qt"
-           upstream="5.10"
-           revision="bc0625e7a03f9f04fdae3a89137812a216e977cd"
+           upstream="5.11"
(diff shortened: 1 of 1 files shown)


fhajredini:
In that case, would it make more sense to just do a merge request from topic/qt5.11 to master and sumo? @Gagi2k

TSAnsari:
@Gagi2k yes cherry-picking commits from topic/qt5.11 to master will be the next task.

Gagi2k:
The main problem is, that for updating to qt5.11 you need to have a recent version of qtapplicationmanager and for that meta-boot2qt needs to be updated as well. If you update meta-boot2qt to the latest version you will get neptune3-ui and the neptune2-ui recipe got removed. But meta-pelux have bbappends for appman and neptune, so it's likely that you will run into problems. That's why recommendation is to keep the effort low and simply pick what has already been done and get 5.11 + neptune3 with low effort ;-)

fhajredini:
https://github.com/lgtmco/lgtm 

This was my first result on google lol.

Merging.

PR #174, opened 2018-07-06
Reverting changes

fhajredini:
Changes being reverted due to errors on built with qt5.11 .

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -30,7 +30,7 @@
            path="sources/meta-bistro" />
 
   <project remote="github"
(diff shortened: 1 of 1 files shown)


sashko:
so, would you like to merge them?

fhajredini:
The only reason i reverted was due to the bug, im working on it to find a solution.

fhajredini:
Apparently the error "internal compiler error killed (program cc1plus)" was related to my environment. The built had no failures on other environments therefore i'm closing this request.

PR #175, opened 2018-07-18
WIP: Add rpm to the test cases

fhajredini:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -26,7 +26,7 @@ void buildNightlyOrWeekly(String variant, String image, boolean checkSmokeTests 
     boolean nightlyBuild = env.NIGHTLY_BUILD == "true"
     boolean weeklyBuild = env.WEEKLY_BUILD == "true"
 
-    if (nightlyBuild || weekly) {
+    if (nightlyBuild || weeklyBuild) {
         buildOnYoctoNode(variant, image, checkSmokeTests)
     } else {
         println("Nothing to do for " + variant)
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -131,6 +131,27 @@ void runSmokeTests(String yoctoDir, String imageName) {
     }
 }
 
+void runBitbakeTests(String yoctoDir) {
+    String archiveDir = "bitbake_TestReport"
+
+    try {
+        stage("Perform Bitbake Testing"){
+            vagrant("/vagrant/cookbook/yocto/runqemu-bitbake-test.sh)
+        }
+    } catch (e){
+	   echo "Bitbake tests failed"
+	   println(e.getMessage())
+    } finally{
+        stage("Publish smoke test results") {
+            reportsDir="/vagrant/${archiveDir}/test_reports/bitbake_tests/"
+            vagrant("mkdir -p ${reportsDir}")
+            vagrant("cp -a ${yoctoDir}/build/TestResults* ${reportsDir}")
+            junit "${archiveDir}/test_reports/bitbake_Tests/TestResults*/*.xml"
+        }
+    }
+}
+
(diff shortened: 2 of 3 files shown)


fhajredini:
the Rpm test suite contains `RpmBasicTest` and `RpmInstallRemoveTest` classes.  However `rpm.RpmInstallRemoveTest.test_check_rpm_install_removal_log_file_size` test fails. Therefore this test is not included.

sashko:
Why does the test fail then?

fhajredini:
Failed 
rpm.RpmInstallRemoveTest.test_check_rpm_install_removal_log_file_size (from rpm.RpmInstallRemoveTest-20180720135533)

Error Message
invalid literal for int() with base 10: 'grep: /var/log/messages: No such file or directory\n0'

Stacktrace

Traceback (most recent call last):
File "/home/yoctouser/pelux_yocto/build/../sources/poky/meta/lib/oeqa/core/decorator/__init__.py", line 32, in wrapped_freturn func(*args, **kwargs)

File "/home/yoctouser/pelux_yocto/build/../sources/poky/meta/lib/oeqa/core/decorator/__init__.py", line 32, in wrapped_freturn func(*args, **kwargs)

File "/home/yoctouser/pelux_yocto/sources/poky/meta/lib/oeqa/runtime/cases/rpm.py", line 141, in test_check_rpm_install_removal_log_file_sizeself.assertGreaterEqual(int(output), 80,

ValueError: invalid literal for int() with base 10: 'grep: /var/log/messages: No such file or directory\n0'

Jenkinsfile
```diff
@@ -8,36 +8,39 @@
 // that one has to allocate a node first, which would stay busy until the end of
 // the parallel pipeline
 
-void buildOnYoctoNode(String variant, String image, boolean checkSmokeTests = false) {
+void buildOnYoctoNode(String variant, String image, boolean checkSmokeTests = false, boolean checkBitbakeTests=false) {
```
thenor: I think I had issues before when I tried to use more than one optional. If there is one out of two unset variables in a row it should be impossible for the script to know which one is not set.

ci-scripts/yocto.groovy
```diff
@@ -216,8 +236,15 @@ void buildManifest(String variantName, String imageName, boolean smokeTests=fals
             boolean buildUpdate = variantName.startsWith("rpi")
             buildImageAndSDK(yoctoDir, imageName, buildUpdate)
             if (smokeTests) {
-                runSmokeTests(yoctoDir, imageName, testBitbake)
+                boolean weekly = env.WEEKLY_BUILD == "true"
+		runSmokeTests(yoctoDir, imageName)
+		if(weekly) {
+		    run runBitbakeTests(yoctoDir)
+		}				
+
```
thenor: Yes, this is what I was thinking of as a solution.
The indentation and whitespace looks a bit strange though.

fhajredini: ok. Will fix it once im back. I was working on vim so didnt notice it.

ci-scripts/yocto.groovy
```diff
@@ -131,6 +131,26 @@ void runSmokeTests(String yoctoDir, String imageName) {
...
+	stage("Perform Bitbake Testing"){
+	    vagrant("/vagrant/cookbook/yocto/runqemu-bitbake-test.sh)
+	}
+    } catch (e){
+	echo "Bitbake tests failed"
+	println(e.getMessage())
+    } finally{
+	stage("Publish smoke test results") {
+	    reportsDir="/vagrant/${archiveDir}/test_reports/bitbake_tests/"
+            vagrant("mkdir -p ${reportsDir}")
```
thenor: The indentation looks a bit strange here also.

PR #177, opened 2018-07-27
pelux.xml: reorganize repos by category

sashko:


TSAnsari:
What are the categories here?

sashko:
Base (poky, oe), Pelagicore/Luxoft, GENIVI, 3rd-party, BSP layers.

PR #187, opened 2018-08-09
pelux.xml: bump meta-pelux

sashko:


TSAnsari:
Since you are soon reverting this commit in https://github.com/Pelagicore/meta-pelux/pull/169, why need to bump to this revision?

sashko:
@TSAnsari good point, I'll update the patch.

PR #192, opened 2018-08-26
local.conf.appendix: add local source mirror

sashko:
Added local source mirror to use as a temporary fix when other layers
fix broekn SRC_URI in own recipes.


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "cookbook"]
 	path = cookbook
-	url = https://github.com/Pelagicore/vagrant-cookbook.git
+	url = https://github.com/sashko/vagrant-cookbook.git
```
thenor: Don't forget to change this before merging.

PR #196, opened 2018-08-29
site.conf.in:fix cache

fhajredini:
the file:// added is a protocol that looks for files locally.
Previously when u wanted to build with cache you had to use "file:///var/yocto-cache" now you only type: "/var/yocto-cache" 


fhajredini:
Closed because we might want to use `ftp` for external mirroring.

kanavin:
I have to say: don't use ftp :) It's basically dead as a protocol, there are no actively supported server implementations, Yocto project has obsoleted it for SRC_URI, Debian has done the same etc.
https://www.debian.org/News/2017/20170425

fhajredini:
Yet, we might use it for other protocols. Ftp would most probably not be used.

PR #231, opened 2018-09-27
README: fix new line break in copyright

sashko:


sashko:
@FlorentRevest you were too fast. I just wanted to see how it'll look on github first and once confirmed that the fix works, I've updated the commit message :)

FlorentRevest:
Ahah, no worries then.

PR #234, opened 2018-10-01
Fix intermittent IO errors in CI/CD by moving Yocto build dir to bind mount from overlay2 fs

martin-ejdestig:
Have run a couple of full builds (intel, intel-qtauto, rpi and rpi-qtauto) on stockholm, reykjavik, kyiv and helsinki with this and all intermittent build errors are gone. Did not manage to pass a single full build on any of these nodes before.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,9 @@ FROM crops/yocto:ubuntu-16.04-base
 
 LABEL description="PELUX Yocto build environment"
 
-# Enables us to overwrite the user ID for the yoctouser. See below
+# Enables us to overwrite the user and group ID for the yoctouser. See below
 ARG userid=1000
+ARG groupid=1000
 
 USER root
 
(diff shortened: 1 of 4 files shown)


rpannek:
You're removing `checkout scm`, how do you checkout the pelux-manifests git repo during a nightly build which is not triggered by a different job like meta-pelux?

martin-ejdestig:
@rpannek It is already done in JenkinsFile that then imports ci-scripts/yocto.groovy. I wrote a sentence about that in the commit message but perhaps I should clarify a bit.

thenor:
The title of this pull request is only reflecting one of the commits. I guess the total change is larger than just that one commit. I would prefer a more explanatory PR title.

martin-ejdestig:
The commits prior to the last one are cleanups (that make the final commit "simpler") or fixes (required for final commit to work "more correctly"). Not sure what to put in the title. Do not want to squash commits (small commits, that makes sense on their own, ftw!).

rpannek:
I read the commit message but didn't quite get that from it, thanks for clearifying.

PR #236, opened 2018-10-02
pelux.xml: bump bunch of layers

sashko:
Updated poky and meta-openembedded to the latest revisions, since both
contain lots of security fixes.

Newer meta-boot2qt has adaptations to the latest poky and oe, which
got some packages upgraded.

meta-swupdate just contains a couple of minor fixes.

Bumped meta-bistro and meta-pelux as well, because why not.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -13,13 +13,13 @@
   <!-- Base stuff -->
   <project remote="yocto"
            upstream="sumo"
-           revision="45ef387cc54a0584807e05a952e1e4681ec4c664"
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
Please rebase to latest sumo to get CI/CD fixes.

martin-ejdestig:
Build probably failed because it is based on sumo and sumo is not sync:ed with master (meta-pelux is missing qtapplicationmanager fixes etc).

PR #247, opened 2018-10-08
Archive images in Jenkins

thenor:
Archive images for ARP and Intel when building with Jenkins.

### DIFF
--- a/ci-scripts/copy_to_archive
+++ b/ci-scripts/copy_to_archive
@@ -28,20 +28,20 @@ ARCHIVE_DIR="$2"
 IMAGE_DIR="${BUILD_DIR}/tmp/deploy/images"
 SDK_DIR="${BUILD_DIR}/tmp/deploy/sdk"
 
-RPI_ARCH="raspberrypi3"
 INTEL_ARCH="intel-corei7-64"
+ARP_ARCH="arp"
 
 IMAGE_BASENAME="core-image-pelux"
 SDK_BASENAME="pelux-glibc"
 
-# If there is a rpi directory available
-if [ -d ${IMAGE_DIR}/${RPI_ARCH} ]; then
-    echo "Archiving rPi image to $ARCHIVE_DIR"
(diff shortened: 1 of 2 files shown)


ci-scripts/copy_to_archive
```diff
@@ -30,22 +30,30 @@ SDK_DIR="${BUILD_DIR}/tmp/deploy/sdk"
...
 IMAGE_BASENAME="core-image-pelux"
 SDK_BASENAME="pelux-glibc"
 
+##### RPI image storing is disabled for now to save capacity #####
 # If there is a rpi directory available
-if [ -d ${IMAGE_DIR}/${RPI_ARCH} ]; then
-    echo "Archiving rPi image to $ARCHIVE_DIR"
-    cp ${IMAGE_DIR}/${RPI_ARCH}/${IMAGE_BASENAME}-*.rpi-sdimg $ARCHIVE_DIR
-fi
+#if [ -d ${IMAGE_DIR}/${RPI_ARCH} ]; then
```
sashko: why leaving this commented?

thenor: In case someone wanted to enable it easily, but I guess that could also be done using the git history.

ci-scripts/yocto.groovy
```diff
@@ -228,7 +228,8 @@ void buildManifest(String variantName, String imageName, boolean smokeTests=fals
             archiveCache(yoctoDir, archive, env.YOCTO_CACHE_ARCHIVE_PATH)
             // If nightly build, we store the artifacts as well
             boolean nightly = env.NIGHTLY_BUILD == "true"
-            if (nightly) {
+            boolean weekly = env.WEEKLY_BUILD == "true"
+            if (nightly || weekly) {
```
fhajredini: Based on this, the arp image will be archived once a week, while all the other images will be archived 5 times a week. Are we okay with this?

thenor: My reasoning here is that for now this could be ok. ARP isn't updated very often and we can change which variants are included in the Nightly/Weekly very easily. ARP was put in the weekly originally since it is not updated very often, but at this moment we aren't including any other variants, but I think we should.

PR #254, opened 2018-10-12
test-scripts/local.conf: remove auto tests temporarily

fhajredini:
Test scripts have been failing because of a package (galculator) being unable
to be tar-ed. Its format is bz2 but neither bzip2, or lbzip2 was installed
on the target. This is not a fix, this is a temporarily workaround.


sashko:
have you tried excluding `gcalculator` script?

PR #261, opened 2018-10-24
Dockerfile: upgrade image to Ubuntu 18

sashko:


thenor:
Why are you removing the pull command? The image on Docker hub should be updated instead?

ci-scripts/yocto.groovy
```diff
@@ -21,12 +21,6 @@ void startVagrant(boolean kill=true) {
             shutdownVagrant()
         }
 
-        // Pull Docker container from Docker Hub
-        sh "docker pull pelagicore/pelux_ubuntu1604"
-
-        // Tag container as 'pelux'
-        sh "docker tag pelagicore/pelux_ubuntu1604 pelux"
-
         sh "vagrant up"
```
thenor: Why remove this?

sashko: We've removed usage of the pre-built container to simplify debugging of the CI/CD issues. Once it's stable, we'll update the image on Docker Hub.

thenor: Oh I didn't know this.

martin-ejdestig: @sashko Just put what you wrote here in the commit message. A perfect "why". :) The commit message as it is now only contains "what". :(

Vagrantfile
```diff
@@ -7,6 +7,7 @@ Vagrant.configure(2) do |config|
         d.has_ssh = true
         d.build_args = ['--build-arg', 'userid=' + `id -u`.strip, '--build-arg', 'groupid=' + `id -g`.strip]
         d.create_args = ['--cap-add=NET_ADMIN', '--device=/dev/net/tun']
+        d.create_args = ['--security-opt', 'seccomp:unconfined']
```
martin-ejdestig: Write details in commit message instead of referring to a link that you have no control over? Maybe even leave a comment in Vagrantfile?

martin-ejdestig: Can of course keep link as well, but not only link. Might be gone tomorrow because SO changed their db or something for all we know.

sashko:
@martin-ejdestig ping

Vagrantfile
```diff
@@ -7,6 +7,7 @@ Vagrant.configure(2) do |config|
         d.has_ssh = true
```
fhajredini: On the 3rd line, i believe this should be added: 
` ENV['VAGRANT_DEFAULT_PROVIDER'] = 'docker' `
Tested qemu tests with this and the taps are being successfully created.

fhajredini:
I have just tested this again, and it works as it is.

thenor:
Since we can't get the automated (qemu) tests to work with Ubuntu 18 and yocto actually doesn't include 18 in their list of supported Linux distributions we will close this. Might be opened again when yocto supports Ubuntu 18.

PR #262, opened 2018-10-31
Run yocto compatibility check weekly

TSAnsari:


martin-ejdestig:
Will the weekly fail until we fix all errors now?

thenor:
Is there a way to make this not be a fatal issue for a build if not all tests pass? E.g. make it yellow (unstable).

TSAnsari:
I am not sure how to make it non-fatal. Do you know how to do that @thenor . At the moment I am trying to see if I can fix the errors that show up.

martin-ejdestig:
vagrant() has a saveStatus argument that I am guessing you can use.

I also think that sh will throw by default if the exit code is not 0 (which is how the script you are calling communicates error, right?), so you could just try/catch in runYoctoCheckLayer() where you call vagrant().

Regardless of how you handle it I think it would be a good idea to do a custom test run on Jenkins before merging since the weekly is run so seldom.

TSAnsari:
Thanks @martin-ejdestig I will try that solution

PR #263, opened 2018-10-31
pelux.xml: switch to thud

sashko:


thenor:
Will there be an update of this PR or should this be closed and a new one opened with a new base?

PR #269, opened 2018-11-08
Pelux 3.0 bump

thenor:


thenor:
I would like to test this manually on at least one target so neptune-UI is still working as expected.

thenor:
This is not building, so I will postpone this work until next week because this week there are more urgent tasks.

fhajredini:
This PR is gonna be closed i assume?

thenor:
@fhajredini I think Oleksandr may have done most of this work already, so yes.

thenor:
We will do this work later. No need to keep this PR open.

PR #276, opened 2018-11-14
yocto.groovy: Run compatibility tests on every build

TSAnsari:


sashko:
on every build?

TSAnsari:
@sashko Isn't this what we said in the daily standup, since weekly and nightly pipelines are not running properly due to global configs overriding the local pipeline configs in Jenkins, resulting in these checks not running. So we add it where it run even with global config.

thenor:
@TSAnsari I don't know if that is what was said, but that might be one way to work around the issue.

sashko:
I'm sorry, but I don't like the idea of running additional code we need to look very occasionally into on every single build.

PR #279, opened 2018-11-15
Do not build -dev images automatically

fhajredini:
When you build the -dev image automatically, it means that
when you try to actually build a dev image, it will go on the
update image as well and do the "-dev-dev-update". This is
inconsistent with the other builds that we do and whenever we try to
build rpi -dev images, it always fails because of not finding the
specified image.


thenor:
The word "stop" in the commit message implies to me that we stop some process which is started automatically (or in some way that we can't affect), so I would prefer to use the phrase "Do not" or in some other way slightly rephrase the first line.

PR #282, opened 2018-11-15
Cherry-pick Jenkins fixes to sumo

thenor:
Some of these changes are for getting things to work at all, some are to just align better with master.


ci-scripts/yocto.groovy
```diff
@@ -135,6 +132,13 @@ void runSmokeTests(String yoctoDir, String imageName) {
     }
 }
 
+void runBitbakeTests(String yoctoDir) {
```
thenor: I just realized that this might never be run, since we don't ever call the function, but I guess it's nice to include it anyway. I think I got it as part of another PR during the merge or something.

PR #295, opened 2018-11-27
Revert "yocto.groovy: Run compatibility tests on every build"

sashko:
This reverts commit 5c7aad156de24ca91512f1c50179fd78ec153881.

Since virtually all the layers contain compatibility issues which we
can not fix, every build fails due to this check failing.


TSAnsari:
Is it failing because of compatibility check? Compatibility check failure is ignored in the code so it should not have any effect on the build failure.

sashko:
@tsansari yes, check any of the recent builds, please

thenor:
@sashko The yocto part is failing, but as far as I can see those builds are still marked as passed. At least on master there is no issue.
The builds for this PR however seem to fail on not being able to fetch the git-lfs package.

TSAnsari:
@sashko  I noticed the same as @thenor. So the build failure is not due to compatibility check, though the checks themselves are failing.

PR #296, opened 2018-11-29
ci-scripts: remove license headers

sashko:
Added branch parameter to the script to make it possible to clone the
manifest of branches other than master.

### DIFF
--- a/ci-scripts/do_repo_init
+++ b/ci-scripts/do_repo_init
@@ -1,23 +1,7 @@
-#! /bin/bash
+#!/bin/bash
 #
 # Copyright (C) 2017 Luxoft
 #
-# Permission to use, copy, modify, and/or distribute this software for
-# any purpose with or without fee is hereby granted, provided that the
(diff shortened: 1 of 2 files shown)


thenor:
I asked the question to Jeremiah, he said smaller files don't need it, but larger files should contain the header. Even if we already have a License file. So I don't really care about this instance. It just wouldn't bother me to leave it either.

ci-scripts/do_repo_init
```diff
@@ -1,39 +1,30 @@
...
-# For further information see LICENSE
-#
-
 
 MANIFEST="$1"
+BRANCH="$4"
+
+# Parameter expansion with default values
+SYNC_DIR="${2:-/vagrant}"
+YOCTO_DIR="${3:-pelux_yocto}"
```
thenor: This comment was marked as resolved, but I never got an answer to my question.

thenor:
lol this changed into something completely different before the merge.

PR #306, opened 2018-12-17
Jenkinsfile: add a job specific variant list

thenor:
Since the global variant list from the Jenkins env overrides the
job specific list (due to an issue in Jenkins) we add this as a
workaround.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ try {
     println "[\"${variantList.join('", "')}\"]"
 } catch(e) {
     println(e.getMessage())
-    echo "Using the default \"VARIANT_LIST\" for build"
+    echo "Using the default variant list for build"
     variantList = ['intel-qtauto:core-image-pelux-qtauto-neptune-dev',
                     'rpi-qtauto:core-image-pelux-qtauto-neptune-dev']
     println "[\"${variantList.join('", "')}\"]"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -19,7 +19,21 @@ void buildOnYoctoNode(String variant, String image) {
 }
 
 def variantMap = [:]
-def variantList = env.VARIANT_LIST.split()
+def variantList = []
+
+// If VARIANT_LIST is defined in the Jenkins Environment
+// then use that, otherwise use the default
+try {
+    variantList = env.VARIANT_LIST.split()
+    echo "Using the specified variant list for build"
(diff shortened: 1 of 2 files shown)


thenor:
I haven't completed the changes for yocto.groovy, but since this change solves a big part of our issue perhaps we want to merge it before the rest of the change is done.

I also added the possibility to archive even if the build is not Nightly or Weekly.

Jenkinsfile
```diff
@@ -19,7 +19,16 @@ void buildOnYoctoNode(String variant, String image) {
...
 
 def variantMap = [:]
-def variantList = env.VARIANT_LIST.split()
+def variantList = []
+
+try {
+    variantList = env.JOB_VARIANT_LIST.split()
+} catch(e) {
+    println(e.getMessage())
+    echo "Using the default \"VARIANT_LIST\" for build"
```
fhajredini: Maybe specifying here (in a short way such as `rpi-qtauto-dev`) which variants we are building would be better.

ci-scripts/yocto.groovy
```diff
@@ -229,13 +229,25 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
...
         } finally { // Archive cache even if there were errors.
             archiveCache(yoctoDir, archive, env.YOCTO_CACHE_ARCHIVE_PATH)
-            // If nightly build, we store the artifacts as well
+            // Check if we want to store the artifacts as well
+            boolean archiveArtifacts = false
             boolean nightly = env.NIGHTLY_BUILD == "true"
             boolean weekly = env.WEEKLY_BUILD == "true"
-            if (nightly || weekly) {
+            try {
+                archiveArtifacts = env.JOB_ARCHIVE_ARTIFACTS
```
fhajredini: Changing the name for the archiveArtifacts would be nice. We have multiple uses of it.

thenor: Done. Please review the name changes I did. Both for the boolean and the method

Jenkinsfile
thenor: I don't understand this comment. Do you mean that the long string is hard to read? Should I have the default values specified in a comment?

fhajredini: echo "Using \"rpi-qtauto-dev\" and \"intel-qtauto-dev\" for build"

thenor: Ok I get it. Yes I will add it.

PR #308, opened 2019-01-14
ISAFW: Run image security analysis on images

fhajredini:
Integrated to run on our CI/CD environment via a local variable.
When triggered, it will save the logs and release them as artifacts
together with other images.

### DIFF
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -46,7 +46,7 @@ void repoInit(String manifest, String yoctoDir) {
     }
 }
 
-void setupBitbake(String yoctoDir, String templateConf, boolean archive, boolean smokeTests=false) {
+void setupBitbake(String yoctoDir, String templateConf, boolean doArchiveCache, boolean smokeTests=false, boolean analyzeImage=false) {
     stage("Setup bitbake") {
         vagrant("/vagrant/cookbook/yocto/initialize-bitbake.sh ${yoctoDir} ${templateConf}")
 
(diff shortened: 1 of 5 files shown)


ci-scripts/yocto.groovy
```diff
@@ -46,7 +46,7 @@ void repoInit(String manifest, String yoctoDir) {
     }
 }
 
-void setupBitbake(String yoctoDir, String templateConf, boolean archive, boolean smokeTests=false) {
+void setupBitbake(String yoctoDir, String templateConf, boolean archive, boolean smokeTests=false, boolean analyzeImage=false) {
```
thenor: I just want to make sure that this works, since it has two optionals. I believe we have had issues with this before in groovy.

ci-scripts/yocto.groovy
```diff
@@ -166,7 +171,6 @@ void archiveImagesAndSDK(String yoctoDir, String suffix) {
 
         sh "rm -rf ${artifactDir}"
         sh "mkdir ${artifactDir}"
-
         // Copy images and SDK to the synced directory
         vagrant("/vagrant/ci-scripts/copy_to_archive ${yoctoDir}/build /vagrant/${artifactDir}")
```
fhajredini: for the archiveArtifacts, add try/catch so we never get a null pointer exception.

thenor:
This build's automatic build fails, but the same change has been built here: https://pelux.io/jenkins/job/fhajredini-pelux-manifest/22/

PR #310, opened 2019-01-21
pelux.xml: bump pelux revisions

TSAnsari:
Point to latest master revisions:
meta-pelux
meta-bistro
meta-template

Point to latest qt 5.12 revisions:
meta-qt5
meta-boot2qt

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -25,20 +25,20 @@
 
   <!-- Pelagicore/Luxoft stuff -->
   <project remote="github"
-           upstream="sumo"
-           revision="63751568790d2a4025270e03e798804bcfb52a2b"
+           upstream="master"
+           revision="f28fc5c6b070409150472b54b716b1c9c5f5360e"
            name="Pelagicore/meta-bistro"
            path="sources/meta-bistro"/>
 
   <project remote="github"
-           upstream="sumo"
-           revision="27331572026b0a3fb467846d2beee893dd16180d"
+           upstream="master"
+           revision="5b57493057050f0fa8d5f4f793f6912583a860cd"
            name="Pelagicore/meta-pelux"
            path="sources/meta-pelux"/>
 
   <project remote="github"
-           upstream="sumo"
-           revision="f6e35df5da50f97ca003db99e3d99f10e8b695b6"
+           upstream="master"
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -71,13 +71,13 @@
...
            upstream="5.12"
-           revision="6830447387d2bfaef1e855040068df2a7f19d485"
+           revision="b676fea3938221781bf81a8cfcf40c7d2046f8d8"
            name="yocto/meta-qt5"
            path="sources/meta-qt5"/>
 
   <project remote="code.qt"
            upstream="sumo"
-           revision="880ecb7a6e993a986d24c9a8b813bf2a48f51d24"
+           revision="5e9819044b5536eea9ad01cce0a6cc85d8b1ea6b"
```
JEderonn: I would prefer to bump to 5.12 75e93a6ecb121d9aa0ee4aa986b3eef0dd20ea22 so we have a "stable" version.

pelux.xml
```diff
@@ -71,13 +71,13 @@
   <!-- Qt + Qt Automotive support stuff -->
   <project remote="code.qt"
            upstream="5.12"
-           revision="6830447387d2bfaef1e855040068df2a7f19d485"
+           revision="b676fea3938221781bf81a8cfcf40c7d2046f8d8"
```
JEderonn: I would prefer to not bump this since it's already 5.12.

TSAnsari:
The build failed on Jenkins due to gammaray issue. Previously, there was a fix bbappend file added for gammaray in meta-pelux. However, that issue has now been fixed in qt5.12, so the fix in meta-pelux is failing the build. Removing this fix from meta-pelux in PR 
https://github.com/Pelagicore/meta-pelux/pull/241
Once the gammaray fix is merged in meta-pelux, this bump should work fine.

TSAnsari:
So it turns out that we need to make some changes to the meta-pelux neptune.service file, which overrides the neptune.service file in meta-boot2qt layer. Some config file names have changed in newer versions of qt as well as how neptune3 is started, so we need to update meta-pelux service file to reflect that. So  this PR is on hold until I fix the service file in the meta-pelux layer.

PR #316, opened 2019-02-02
pelux.xml: switch to thud

sashko:

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -2,7 +2,7 @@
 <manifest>
 
   <!-- Settings -->
-  <default sync-j="4" revision="master"/>
+  <default sync-j="4" revision="thud"/>
 
   <!-- Remotes -->
   <remote fetch="git://git.openembedded.org" name="oe"/>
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -31,10 +31,10 @@
            path="sources/meta-bistro"/>
 
   <project remote="github"
-           upstream="master"
-           revision="5b57493057050f0fa8d5f4f793f6912583a860cd"
+           upstream="thud"
+           revision="fd6444a7d67f7071b1d7421bad22d97d4957434d"
            name="Pelagicore/meta-pelux"
-           path="sources/meta-pelux"/>
+           path="sources/meta-pelux" />
```
TSAnsari: remove space

pelux.xml
```diff
@@ -31,10 +31,10 @@
            path="sources/meta-bistro"/>
```
TSAnsari: meta-bistro should also be moved to thud.

sashko: yes, I don't wanna branch them atm; I had to branch `meta-pelux` to remove `mopidy`.

thenor: Just to clarify we won't make a branch on an internal layer just to have a branch with a certain name. While there isn't a release to be done soon we will point to master and before the next release we will create the thud branch, but since there are no changes needed to adapt to thud then there is no point in making the branch at this time.

PR #319, opened 2019-02-26
pelux.xml: bump poky

sashko:


thenor:
I clicked Approve, but then I realized you didn't include a list of changes between the commits in the commit message. Is it possible to add this?

PR #327, opened 2019-03-08
Vagrantfile: fix path to Yocto cache

sashko:
YOCTO_CACHE_URL environment variable is used by Jenkins to set path to
the Yocto cache, not the currently specified YOCTO_CACHE_PATH.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -15,8 +15,8 @@ Vagrant.configure(2) do |config|
 
     # If an archive path for the yocto cache is given, we mount it into the vm
     # using the same path as on the host.
-    unless ENV['YOCTO_CACHE_PATH'].to_s.strip.empty?
-      config.vm.synced_folder ENV['YOCTO_CACHE_PATH'], ENV['YOCTO_CACHE_PATH']
+    unless ENV['YOCTO_CACHE_URL'].to_s.strip.empty?
+      config.vm.synced_folder ENV['YOCTO_CACHE_URL'], ENV['YOCTO_CACHE_URL']
     else
       config.vm.synced_folder "/var/yocto-cache", "/var/yocto-cache"
     end


martin-ejdestig:
> YOCTO_CACHE_URL environment variable is used by Jenkins to set path to
the Yocto cache, not the currently specified YOCTO_CACHE_PATH.

This is not correct.

YOCTO_CACHE_URL is used to set CACHEURL ... and is a URL, so should not be used here. Or are you saying that it works to have a URL for config.vm.synced_folder?

YOCTO_CACHE_PATH is used when archiving the cache (and making the URL in YOCTO_CACHE_URL be bind mounted in Vagrantfile so that the file:// URL can actually be read from).

See how the variables are used in ci-scripts/yocto.groovy .

We could refactor this, of course so that there is only a single variable. But I do not know if the original intent was to make it possible to have the cache e.g. be a http(s) URL. No idea.

It is the weekend and I am not supposed to work (curse of having personal and work GH account be the same :sob:). We can check more next week.

sashko:
Alright, I need to actually dig into the code. I can clearly see that cache is **not** used atm, and the following in the log:

```
Trying to fetch string YOCTO_CACHE_URL
[Pipeline] echo
Value for YOCTO_CACHE_URL is file:///var/yocto-cache
[Pipeline] echo
Trying to fetch string YOCTO_CACHE_PATH
 > git rev-parse refs/remotes/origin/ubuntu_18^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/refs/heads/ubuntu_18^{commit} # timeout=10
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 24171611fcfaded261a390aaf1b417725be7610e # timeout=10
 > git rev-list --no-walk 61962ebd51e15737f4c5a9a061699003ad165d21 # timeout=10
[Pipeline] echo
java.lang.Exception: The result of YOCTO_CACHE_PATH was null
```

sashko:
turns out, my test job was cloned from the old version of `pelux-manifest_SMOKE_TEST`

thenor:
Just so you know the " ... was null" is the ugly result of how I had to do the check in Jenkinsfile. It's possible there is a better solution, but some reason I couldn't do it the same way as in the yocto.groovy.

PR #329, opened 2019-03-15
copy-to-archive: copy only the timestamped image

fhajredini:
Currently we copy duplicate images. One with the timestamp
and the other without the timestamp.

This change copies only the one with a timestamp.

Signed-off-by: Fisnik Hajredini <fisnikhajredini@gmail.com>

### DIFF
--- a/ci-scripts/copy_to_archive
+++ b/ci-scripts/copy_to_archive
@@ -39,19 +39,19 @@ fi
 # If there is an Intel directory
 if [ -d ${IMAGE_DIR}/${INTEL_ARCH} ]; then
     echo "Archiving Intel image to $ARCHIVE_DIR"
-    mv ${IMAGE_DIR}/${INTEL_ARCH}/${IMAGE_BASENAME}-*.wic $ARCHIVE_DIR
+    mv ${IMAGE_DIR}/${INTEL_ARCH}/${IMAGE_BASENAME}-*rootfs.wic $ARCHIVE_DIR
 fi
 
 # If there is an ARP directory
 if [ -d ${IMAGE_DIR}/${ARP_ARCH} ]; then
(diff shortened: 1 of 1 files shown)


sashko:
Hold on, isn't the other one is just a symbolic link?

martin-ejdestig:
I would suggest putting something like (copied your text and added a bit):

```
Currently we copy duplicate images. One with the timestamp and the other
without the timestamp. This is due to there being a symlink in the
tmp/deploy/images folder. E.g.:

core-image-pelux-qtauto-neptune-dev-intel-corei7-64-20190307122631.rootfs.wic
core-image-pelux-qtauto-neptune-dev-intel-corei7-64.wic -> core-image-pelux-qtauto-neptune-dev-intel-corei7-64-20190307122631.rootfs.wic

When archiving we actually copy what the symlinks points to so the same
file is copied twice.

This change copies only the one with a timestamp.
```

sashko:
> I would suggest putting something like (copied your text and added a bit):
> 
> ```
> Currently we copy duplicate images. One with the timestamp and the other
> without the timestamp. This is due to there being a symlink in the
> tmp/deploy/images folder. E.g.:
> 
> core-image-pelux-qtauto-neptune-dev-intel-corei7-64-20190307122631.rootfs.wic
> core-image-pelux-qtauto-neptune-dev-intel-corei7-64.wic -> core-image-pelux-qtauto-neptune-dev-intel-corei7-64-20190307122631.rootfs.wic
> 
> When archiving we actually copy what the symlinks points to so the same
> file is copied twice.
> 
> This change copies only the one with a timestamp.
> ```

You should be an advisor to statsminister.

martin-ejdestig:
Sign off line was wrong.

PR #331, opened 2019-03-20
pelux.xml: bump layers for PELUX 3.1

sashko:

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -13,7 +13,7 @@
   <!-- Base stuff -->
   <project remote="yocto"
            upstream="sumo"
-           revision="fa962ec72f8a8337eaae71ab622e340f6ef56395"
+           revision="b28f5672ab55ea303727e9f03bc594c7774d597e"
            name="poky"
            path="sources/poky"/>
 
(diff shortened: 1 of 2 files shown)


pelux.xml
```diff
@@ -13,7 +13,7 @@
   <!-- Base stuff -->
   <project remote="yocto"
            upstream="sumo"
-           revision="fa962ec72f8a8337eaae71ab622e340f6ef56395"
+           revision="b28f5672ab55ea303727e9f03bc594c7774d597e"
```
JEderonn: Why not use fixed/released version, e.g. sumo-19.0.2?

pelux.xml
```diff
@@ -71,14 +71,14 @@
 
   <project remote="code.qt"
            upstream="sumo"
-           revision="880ecb7a6e993a986d24c9a8b813bf2a48f51d24"
+           revision="75e93a6ecb121d9aa0ee4aa986b3eef0dd20ea22"
```
JEderonn: May make sense to use QtAS-5.12.1, but meta-qt5 shall also be bumped with same tag

sashko:
> I would prefer to use hashes which corresponds to tags, and pl make sure meta-qt5 and meta-boot2qt are consistent, i.e. using same Qt-version.

ok

pelux.xml
```diff
@@ -65,20 +65,20 @@
...
            upstream="5.12"
-           revision="6830447387d2bfaef1e855040068df2a7f19d485"
+           revision="b9131d61d3f08669b493b5070811f2c617c8ee9f"
            name="yocto/meta-qt5"
            path="sources/meta-qt5"/>
 
   <project remote="code.qt"
            upstream="sumo"
-           revision="880ecb7a6e993a986d24c9a8b813bf2a48f51d24"
+           revision="c20e297c8a13b37079dc1a03fcc22d0de2bcece7"
```
JEderonn: Sorry for not being clear in my previous comment, but for enhanced readability, consider use tag v5.12.1_QtAS iof hash OR add a comment about it in the commit message, e.g.
Bumped meta-boot2qt to v5.12.1_QtAS
Bumped meta-qt5 to v5.12.1
...

W/o this clarification, it's not easy at all to know which version of the projects the manifest uses.

sashko:
@JEderonn `QtAS-5.12.1` of `meta-boot2qt` and `5.12.1` tag of `meta-qt` are not buildable together.

Would you mind if we've used what works and was tested?

JEderonn:
As already discussed, use the working hash.

PR #339, opened 2019-04-05
pelux.xml: bump meta-pelux

sashko:
- systemd: remove network configuration
- huspell-dicts: add Czech, US English and German
- linux-raspberrypi: enable memory cgroup
- pelux.conf: set DISTRO_VERSION to master
- core-image-pelux: remove duplicate swupdate installs


fhajredini:
Approved!

PR #340, opened 2019-04-08
Dockerfile: update the way we fetch git-lfs

thenor:
The is no version of git-lfs available by default for Ubuntu 16.04 and
the script we use to make it possible to install doesn't always succeed.
This fix means that all the commands will be run in the same context.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,10 +36,9 @@ RUN apt-get update && apt-get install -y \
 RUN pip3 install unittest-xml-reporting
 
 # For git-lfs
-RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
-# Pin version to 2.5.* until a 2.6.* version with a fix
-# for https://github.com/git-lfs/git-lfs/issues/3358 is released.
-RUN apt-get install -y git-lfs=2.5.*
+# The downloaded script is needed since git-lfs is not available per default for Ubuntu 16.04
+RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash && sudo apt-get install -y git-lfs
+
 
 # Remove all apt lists to avoid build caching
 RUN rm -rf /var/lib/apt/lists/*


Dockerfile
```diff
@@ -36,10 +36,10 @@ RUN apt-get update && apt-get install -y \
 RUN pip3 install unittest-xml-reporting
 
 # For git-lfs
-RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
-# Pin version to 2.5.* until a 2.6.* version with a fix
-# for https://github.com/git-lfs/git-lfs/issues/3358 is released.
-RUN apt-get install -y git-lfs=2.5.*
+# git-lfs is not available per default for Ubuntu 16.04
```
sashko: this is slightly confusing: what do you by "per default"? we're pinning version, so I guess, it's not that default after all?

thenor: Sure. I can rephrase this.

thenor: I reworded a bit. Is this better?

Dockerfile
```diff
@@ -36,10 +36,9 @@ RUN apt-get update && apt-get install -y \
 RUN pip3 install unittest-xml-reporting
 
 # For git-lfs
-RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash
-# Pin version to 2.5.* until a 2.6.* version with a fix
-# for https://github.com/git-lfs/git-lfs/issues/3358 is released.
-RUN apt-get install -y git-lfs=2.5.*
+# The downloaded script is needed since git-lfs is not available per default for Ubuntu 16.04
+RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | sudo bash && sudo apt-get install -y --allow-unauthenticated git-lfs=2.6.*
```
martin-ejdestig: Why do you add `--allow-unauthenticated`? The script imports the PGP key.

thenor: I got a warning for some reason that I should use this flag, then I tested it and it failed, so I got the wrong advice. Removed now.

martin-ejdestig:
Not fond of mixing two things in same commit. I would put changing the way we fetch and install in one commit and updating from 2.5.* to 2.6.* in another (has nothing to do with fixing actual problem).  Or why do we update to 2.6 now?

sashko:
> Not fond of mixing two things in same commit. I would put changing the way we fetch and install in one commit and updating from 2.5.* to 2.6.* in another (has nothing to do with fixing actual problem). Or why do we update to 2.6 now?

only if those changes self-sufficient and work separately, but whatever really for me... just make it work.

thenor:
> Not fond of mixing two things in same commit. I would put changing the way we fetch and install in one commit and updating from 2.5.* to 2.6.* in another (has nothing to do with fixing actual problem). Or why do we update to 2.6 now?

I could also remove the version. I seem to get a 2.7 version if I remove the pinning of version completely.
But this does not seem to work anyway...

thenor:
I have some fascinating news. I have been testing this locally and removing pinning of the version seems to work more often than not.

sashko:
> I have some fascinating news. I have been testing this locally and removing pinning of the version seems to work more often than not.

So, using this script locally does not result in 100% success rate either?

thenor:
> > I have some fascinating news. I have been testing this locally and removing pinning of the version seems to work more often than not.
> 
> So, using this script locally does not result in 100% success rate either?

No...

martin-ejdestig:
Only seen failures in the cloud, correct? It has always worked for me on my own computer.

thenor:
> > > I have some fascinating news. I have been testing this locally and removing pinning of the version seems to work more often than not.
> > 
> > 
> > So, using this script locally does not result in 100% success rate either?
> 
> No...

I meant that I see these issues only on our build servers. I haven't seen them when I run on my own computer.

thenor:
I will merge this, even though it doesn't solve the issue. It is still an improvement of the Dockerfile.

PR #341, opened 2019-04-10
yocto.groovy: make yocto compatibility tests optional

sashko:
Yocto compatibility tests produce inaccurate and confusing results, e.g
they looks for dependencies of every existing package in the tested
layer and if it is not found, the test fails, even though the dependency
is present in another layer that is not used by PELUX and therefore not
included in our bblayers.conf.

Tests also do not distinguish between critical and non-critical errors.

Those tests should occasionally be executed locally instead to ensure
PELUX layers are compatible with the Yocto project.

Disable tests to save on build time.

### DIFF
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -255,6 +255,7 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
         boolean doArchiveCache = getBoolEnvVar("ARCHIVE_CACHE", false)
         boolean smokeTests = getBoolEnvVar("SMOKE_TEST", false)
         boolean bitbakeTests = getBoolEnvVar("BITBAKE_TEST", false)
+        boolean yoctoCompatTest = getBoolEnvVar("YOCTO_COMPATIBILITY_TEST", false)
         setupBitbake(yoctoDir, templateConf, doArchiveCache, smokeTests, analyzeImage)
         setupCache(yoctoDir, yoctoCacheURL)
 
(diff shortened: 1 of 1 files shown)


sashko:
> I would prefer a more complicated change where you add a parameter which is default 'false'.

why?

thenor:
We still want to be able to run the tests on pelux-manifests. Adding a parameter/flag would mean that we could trigger this from Jenkins without having to change the script again.

sashko:
> We still want to be able to run the tests on pelux-manifests. Adding a parameter/flag would mean that we could trigger this from Jenkins without having to change the script again.

This test is completely useless.

thenor:
> > We still want to be able to run the tests on pelux-manifests. Adding a parameter/flag would mean that we could trigger this from Jenkins without having to change the script again.
> 
> This test is completely useless.

It was decided to be included and run at least before releases. But I do agree that they should be disabled for most builds.

sashko:
> > > We still want to be able to run the tests on pelux-manifests. Adding a parameter/flag would mean that we could trigger this from Jenkins without having to change the script again.
> > 
> > 
> > This test is completely useless.
> 
> It was decided to be included and run at least before releases. But I do agree that they should be disabled for most builds.

They should be removed completely as it is not possible to extract any useful information from their results.

Those tests should be run locally in a controlled environment with all the layers included, etc once in a while.

thenor:
I would like to discuss a total removal offline and I don't agree with the way it is removed in this PR.

ci-scripts/yocto.groovy
```diff
@@ -262,7 +262,6 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
         try {
             boolean buildUpdate = variantName.startsWith("rpi")
             buildImageAndSDK(yoctoDir, imageName, variantName, buildUpdate)
-            runYoctoCheckLayer(yoctoDir)
```
fhajredini: The method remains.. you are simply not calling it?

martin-ejdestig:
If we never want to run them in the CI/CD more should be removed from the Groovy scripts than just the single line. If we want to make it configurable, that needs to be done. I do not have any strong opinion on this, just pointing out that removal of this single line is not enough either way IMHO.

PR #351, opened 2019-04-15
Dockerfile: remove inheritance of crops/yocto image

sashko:
The Yocto Project recommends using Docker containers from crops for
building Yocto based projects.

With @ee336cd96e we have started using crops/yocto:ubuntu-16.04-base
as a base for our container, which in theory was supposed to make
maintenance of our Dockerfile eaiser.

On practice, Docker container from crops happened to have lots
unnecessary packages installed, e.g. Fluxbox and VNC server.

By removing inheritance of crops/yocto:ubuntu-16.04-base we make our
image more lightweight and gain more control over its content.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM crops/yocto:ubuntu-16.04-base
+FROM ubuntu:16.04
 
 LABEL description="PELUX Yocto build environment"
 
@@ -10,24 +10,48 @@ USER root
 
 # Install dependencies in one command to avoid potential use of previous cache
 # like explained here: https://stackoverflow.com/a/37727984
-RUN apt-get update && apt-get install -y \
-        openssh-server \
-        iptables \
-        cvs \
+RUN apt-get update && \
+    apt-get install -y \
+        build-essential \
+        chrpath \
         coreutils \
-        python3-pip \
-        libfdt1 \
-        python-pysqlite2 \
-        help2man \
-        libxml2-utils \
-        libsdl1.2-dev \
-        graphviz \
-        qemu-user \
-        g++-multilib \
+        cpio \
         curl \
+        cvs \
+        diffstat \
+        g++-multilib \
+        gawk \
+        gcc-multilib \
+        git-core \
+        graphviz \
+        help2man \
+        iptables \
+        iputils-ping \
+        libfdt1 \
+        libsdl1.2-dev \
+        libxml2-utils \
+        locales \
+        m4 \
+        openssh-server \
+        python \
+        python-pysqlite2 \
+        python3 \
+        python3-pip \
+        qemu-user \
         repo \
         rsync \
-        m4
+        screen \
+        socat \
+        subversion \
+        sudo \
+        sysstat \
+        texinfo \
+        tmux \
+        unzip \
+        wget \
+        xz-utils
+
+RUN apt-get clean
 
 # For Yocto bitbake -c testimage XML reporting
 RUN pip3 install unittest-xml-reporting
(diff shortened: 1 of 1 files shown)


pookimon:
Also we can limit number of RUN calls in Dockerfile, as every call is separate layer which means our image becomes larger. Using && between commands we can at least decrease this number.

sashko:
> Also we can limit number of RUN calls in Dockerfile, as every call is separate layer which means our image becomes larger. Using && between commands we can at least decrease this number.

Aight, how big of performance improvement would this be? Could you link us to any relevant info?

pookimon:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Dockerfile
```diff
@@ -36,25 +64,33 @@ RUN pip3 install unittest-xml-reporting
...
+ARG groupid=1000
+
 # Make sure the user/groupID matches the UID/GID given to Docker. This is so that mounted
 # dirs will get the correct permissions
 RUN usermod --uid $userid yoctouser
 RUN groupmod --gid $groupid yoctouser
 RUN echo 'yoctouser:yoctouser' | chpasswd
 RUN echo 'yoctouser ALL=(ALL) NOPASSWD:SETENV: ALL' > /etc/sudoers.d/yoctouser
 
+USER yoctouser
```
thenor: This is done here on line 86 and on line 91. Does it need to be done twice?

Dockerfile
```diff
@@ -1,33 +1,61 @@
...
+        texinfo \
+        tmux \
+        unzip \
+        wget \
+        xz-utils
+
+RUN apt-get clean
+
+# Remove all apt lists to avoid build caching
+RUN rm -rf /var/lib/apt/lists/*
```
martin-ejdestig: Why did you move this? Cleanup? Separate commit if "yes".

Dockerfile
```diff
@@ -36,25 +64,32 @@ RUN pip3 install unittest-xml-reporting
...
+RUN /usr/sbin/locale-gen en_US.UTF-8
 RUN localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8
 RUN echo 'export LC_ALL="en_US.UTF-8"' >> /etc/profile
 ENV LANG en_US.utf8
 
+RUN useradd -U -m yoctouser
+
+# Enables us to overwrite the user and group ID for the yoctouser. See below
+ARG userid=1000
+ARG groupid=1000
```
martin-ejdestig: Why did you move this? (I guess intent of original author was probably to have ARG at top. Do not know for sure nor what best practice. Regardless, unnecessary to move them if not needed. If it is a cleanup it should be a separate commit.)

martin-ejdestig:
> > Also we can limit number of RUN calls in Dockerfile, as every call is separate layer which means our image becomes larger. Using && between commands we can at least decrease this number.
> 
> Aight, how big of performance improvement would this be? Could you link us to any relevant info?

Please do this in a separate commit if you do it in this pr.

sashko:
> Why do you add:
> 
> * debianutils
> * libegl1-mesa
> * python3-git
> * python3-jinja2
> * python3-pexpect
> 
> in this commit? Perhaps I missed something else when comparing. Can not see them in the crops Dockerfile nor in the `build-install-dumb-init.sh` script.
> 
> If you added them because we need them for something and they are not in the crops image, it should be a separate commit.

It's dependencies for `Warrior`, but yes, I could add them with a separate commit.

Dockerfile
sashko: It just made more sense for me that we cleanup cache once we're done with using `apt-get`.

martin-ejdestig: You are not done with apt-get at this point though. So not only is it wrong, it is hidden together with an unrelated change. Please make separate commits for things like this.

sashko:
Just to be clear: I have created clones of our nightly jobs and ran them on the updated container to make sure it does not break anything.

PR #358, opened 2019-05-08
Jenkinsfile: destroy vagrant container and abort PR build

pookimon:
destroy vagrant container even in case of abortion 
abort PR build if PR was updated

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,10 +13,38 @@ void buildOnYoctoNode(String variant, String image) {
 
         checkout scm
         def manifests = load "ci-scripts/yocto.groovy"
-
-        manifests.buildManifest(variant, image)
+        try {
+            manifests.buildManifest(variant, image)
+        } catch (e) {
+            // This part will be run in case of fail
+            manifest.shutdownVagrant()
+            throw e
+        } finally {
+            // Can be redundant, so just to doublecheck
+            manifest.shutdownVagrant()
+        } 
     }
 }
+// If there is a build for this PR in progress, abort it
+script {
+    def jobname = env.JOB_NAME
+    def buildnum = env.BUILD_NUMBER.toInteger()
+    def job = Jenkins.instance.getItemByFullName(jobname)
+
+    for (build in job.builds) {
+        if (!build.isBuilding()) {
+            continue;
+        }
+        if (buildnum == build.getNumber().toInteger()) {
+            continue;
+        }
+
+        println("Aborting previous running build #${build.number}...")
+
+        build.doStop();
+    }
+}
+
 
 def variantMap = [:]
 def variantList = []
--- a/hooks/build
+++ b/hooks/build
@@ -5,6 +5,6 @@ GROUPID=1003
 
 echo "------ DOCKER HUB BUILD HOOK - START -------"
 
-docker build --build-arg BUILD_DATE=`date -u +"%Y-%m-%d %H:%M:%S"` --build-arg userid=$USERID --build-arg groupid=$GROUPID -t $IMAGE_NAME .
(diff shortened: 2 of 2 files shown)


Jenkinsfile
```diff
@@ -13,11 +13,39 @@ void buildOnYoctoNode(String variant, String image) {
...
         def manifests = load "ci-scripts/yocto.groovy"
+        try {
+            manifests.buildManifest(variant, image)
+        } catch (e) {
+            // This part will be run in case of fail
+            manifest.shutdownVagrant()
+            throw e
+        } finally {
+            // Can be redundant, so just to doublecheck
+            manifest.shutdownVagrant()
```
martin-ejdestig: I do not think this will call shutdownVagrant() (or the extra added one above) in more cases when build is aborted than before. There is a `finally` block with shutdownVagrant() already in buildManifest(). (It will however call it more than necessary for sure if build is not aborted.)

Jenkinsfile
```diff
@@ -13,11 +13,39 @@ void buildOnYoctoNode(String variant, String image) {
...
+        } 
+    }
+}
+// If there is a build for this PR in progress, abort it
```
martin-ejdestig: You probably want this in a separate commit.

hooks/build
```diff
@@ -5,6 +5,6 @@ GROUPID=1003
 
 echo "------ DOCKER HUB BUILD HOOK - START -------"
 
-docker build --build-arg BUILD_DATE=`date -u +"%Y-%m-%d %H:%M:%S"` --build-arg userid=$USERID --build-arg groupid=$GROUPID -t $IMAGE_NAME .
+docker build --build-arg userid=$USERID --build-arg groupid=$GROUPID -t $IMAGE_NAME .
```
martin-ejdestig: What is going on here?

Jenkinsfile
pookimon: I've checked it on my personal repo and it worked
https://pelux.io/jenkins/job/dm_testing/job/master/63/console

hooks/build
pookimon: It is my previous change which is not deserves for separate commit. It redefines build command for dockerhub and I forgot to clean up one extra argument

Jenkinsfile
martin-ejdestig: I was under the impression that Jenkins sent `SIGINT` (or `SIGTERM?) to the command started with `sh` and then just aborted the script when user pressed abort again. I have not tested this though. @tahatrue did most of the experimenting here.

If the script actually continues even when forcefully quitting (user presses "really stop now" or whatever it says) and sends `SIGINT` (or some other signal) we probably want to put this closer to the `sh` call that actually does `vagrant shutdown`. Perhaps even add a loop in `shutdownVagrant()` that that tries until it succeeds or a limit has been reached.

martin-ejdestig: In fact, there is already a `retry(5)` in shutdownVagrant(). Does that not work? Is a try/catch/finally still needed inside of it?

Would be nice to find out how it actually works from some Jenkins documentation and how to make sure that a `sh` really should not be aborted.

I found https://gist.github.com/datagrok/dfe9604cb907523f4a2f after a quick search. `sh` and `retry` documentation should be somewhere in https://jenkins.io/doc/book/pipeline/syntax/ ? Can not find it after a quick glance. :/

hooks/build
martin-ejdestig: If it is not related with other changes, it should be a separate commit (and e.g. mention clearly in the commit message that you forgot it in <hash of original commit>). You can, if you want to, take it in this PR to skip the overhead of creating another PR. Just split it out in a separate commit.

PR #359, opened 2019-05-09
Dockerfile: destroy container after 6 hours

pookimon:
In case of force abort of jenkins job container
remains up and running until next run on this node.
Overnight docker-prune job cannot clean it up as it is not stopped.
This fix will help us to release capacity for other jobs


sashko:
explanation, please.

PR #361, opened 2019-05-14
pelux.xml: bump meta-bistro and meta-pelux

sashko:


pelux.xml
```diff
@@ -26,7 +26,7 @@
   <!-- Pelagicore/Luxoft stuff -->
   <project remote="github"
            upstream="master"
-           revision="16dbb71e44de965e0f6f0b87877efa861ad5afe9"
+           revision="9fe536c4a046270328d98ef55a4c1779ed663ccc"
```
TSAnsari: Would be good to include the latest revision just added
97967697c8a77a944375de745d6179e066302d07
So don't have to bump again

PR #367, opened 2019-05-21
yocto.groovy: Remove catching while building

fhajredini:
The catch was added to understand an error. With this catch on, the
builds will always be successful/green and finding a failed build
will be hard.

### DIFF
--- a/ci-scripts/yocto.groovy
+++ b/ci-scripts/yocto.groovy
@@ -260,9 +260,6 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
                 }
             }
 
-        }  catch(e) {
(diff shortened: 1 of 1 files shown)


sashko:
can't you do something like return -1 in catch instead?

fhajredini:
> can't you do something like return -1 in catch instead?

I'm not sure, its not used anywhere on our groovy script and i thought that this is a simpler solution. You suggest i give it a try?

sashko:
> > can't you do something like return -1 in catch instead?
> 
> I'm not sure, its not used anywhere on our groovy script

Why does it work with `runSmokeTests` then?

PR #370, opened 2019-05-24
Thud new test format

mardy:
Since Yocto `thud`, poky's test report is delivered as a non standard JSON file, which Jenkins and other tools don't understand. For this reason, add a script to convert it into jUnit files.

While the script has been successfully tested in another (proprietary) project, the groovy parts are an adaptation of what I have in the other project. I couldn't test them and I'm not sure when I'll get the time to properly do it. I will eventually get to it, but if you have a machine ready to test this, that would be much faster :-)


fhajredini:
Merging it now.

PR #371, opened 2019-05-27
Vagrant to docker migration

pookimon:
In current configuration custom image and custom yocto.groovy script will be used
to not affect other projects. The main purpose - to get rid of vagrant layer of abstraction
at all and leave native docker as containerization provider. Vagrant directory was replaced
with workspace and all vagrant ssh call were replaced by sh


sashko:
is this wip?

pookimon:
> is this wip?

Yes

PR #376, opened 2019-06-04
 yocto.groovy: port json2junit fix

sashko:
Fix for unit tests made in @8207109c4c41c was lost during the Vagrant
to Docker migration in @bc75e1e506dedb.


pookimon:
Replace "vagrant" with "workspace" and everything should be fine

PR #389, opened 2019-06-26
hooks/build: pass UID/GID of user that builds container

sashko:
Instead of having hardcoded user and group IDs, pass IDs of the user
container is build under.

### DIFF
--- a/hooks/build
+++ b/hooks/build
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 
-USERID=1003
-GROUPID=1003
+USERID=$(id -u)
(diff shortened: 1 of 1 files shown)


pookimon:
It will not work this way. Building is performed on doсkerhub servers and we don't even have idea under which user. To have personalized image you need build it by yourself

sashko:
> It will not work this way. Building is performed on doсkerhub servers and we don't even have idea under which user. To have personalized image you need build it by yourself

Well, if that is the case, this change is in fact useless.

PR #392, opened 2019-06-28
build-in-docker.sh: script for building image inside docker

pookimon:
To simplify building process with docker container

### DIFF
--- a/build-in-docker.sh
+++ b/build-in-docker.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+programname=$0
+function usage {
+    echo "usage: $programname -i image -v variant [-b branch]"
+    echo "   -i image    specify image name would you like to build"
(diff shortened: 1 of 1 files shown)


build-in-docker.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "usage: $programname -i image -v variant [-b branch]"
+    echo "   -i image    specify image name would you like to build"
+    echo "   -v variant  specify variant of the target platform"
+    echo "   -b branch   specify branch name which you would like to use for sources (optional)"
+    echo "   -h          display help"
+
+
+}
+BRANCH='master'
+while getopts hi:v:b: option; do
```
fhajredini: Not sure but it looks like you have missed a ":" ``on hi:v:b:``

pookimon: It's not. ":" is for parameters, which receive argument afterwards. "h" stands for "help"

PR #394, opened 2019-06-28
copy_to_archive: copy bmap file

fhajredini:

### DIFF
--- a/ci-scripts/copy_to_archive
+++ b/ci-scripts/copy_to_archive
@@ -39,6 +39,7 @@ fi
 # If there is an Intel directory
 if [ -d ${IMAGE_DIR}/${INTEL_ARCH} ]; then
     echo "Archiving Intel image to $ARCHIVE_DIR"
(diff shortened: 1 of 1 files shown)


sashko:
why don't I have `wic.bz2` image?

ci-scripts/copy_to_archive
```diff
@@ -47,6 +48,7 @@ fi
 
 if [ -d ${IMAGE_DIR}/${ARP_ARCH} ]; then
     echo "Archiving ARP image to $ARCHIVE_DIR"
+    mv ${IMAGE_DIR}/${INTEL_ARCH}/${IMAGE_BASENAME}-*rootfs.wic.bz2 $ARCHIVE_DIR
```
martin-ejdestig: INTEL_ARCH -> ARP_ARCH

fhajredini:
After discussing with Oleksi*, i think that including the bz2 is not worthy as it will only take more time to archive it and take more space. For now I will only add the bmap file.

The compressed image shall be added when a decision to modify the muxpi script is taken. In that way, we can remove the raw image, and only archive the compressed image.

fhajredini:
ping @sashko @martin-ejdestig

PR #411, opened 2019-07-30
yocto2.groovy: sstate cache copying maintenance

pookimon:
Add mounted/not mounted check
Add progress to rsync
Omit skip messages from rsync
Do not fail build when rsync failed

### DIFF
--- a/ci-scripts/yocto2.groovy
+++ b/ci-scripts/yocto2.groovy
@@ -131,10 +131,20 @@ void runYoctoCheckLayer(String yoctoDir) {
 
 
 void archiveCache(String yoctoDir, boolean doArchiveCache, String yoctoCacheArchivePath) {
-    if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
-        stage("Archive cache") {
-            sh "rsync -trpgO ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    def sC = sh script:"test -d ${yoctoCacheArchivePath}/sstate-cache && test -d ${yoctoCacheArchivePath}/downloads", returnStatus:true
+    if (sC != 0) {
+        echo "Cache dirs are not mounted"
+    } else {
+        echo "Cache dirs are mounted"
+        if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
+            stage("Archive cache") {
+                try {
+                    sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
+                    sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+                } catch(e) {
+                    println("Error archiving cache \n" + e)
+                }
+            }
         }
     }
 }


ci-scripts/yocto2.groovy
```diff
@@ -66,7 +66,14 @@ void buildImageAndSDK(String yoctoDir, String imageName, String variantName, boo
     } else {
         echo "\'site.conf\' exists, using cache"
     }
-
+    stage("Verify NFS") {
+        def sC = sh script:"mount | grep -w nfs4", returnStatus:true
```
martin-ejdestig: Would it not be better to check path here instead of "any NFS fs is mounted"?

martin-ejdestig: Or perhaps just check if the "downloads" folder and "sstate-cache" folder exist. Don't involve "mount" at all. Script is even less specific to our particular setup.

pookimon: Do you mean "mount | grep -w '/var/yocto-cache'"?

pookimon: > Or perhaps just check if the "downloads" folder and "sstate-cache" folder exist. Don't involve "mount" at all. Script is even less specific to our particular setup.

It's not thee best idea, as those directories can be presented locally. Another option is to fail build  if NFS is not mouned (read "dirs sstate-cache and downloads are absent")

martin-ejdestig: Yes, that would make it so this test does not give a false positive if another NFS mount exists on the system.

But also, why require a mount at all. If one wants to not use a mount (NFS or other one, the way we had it before NFS was setup) for some reason, using mount to check would make this warning always trigger. So why not just do what I mentioned in second comment? I.e. just check that  `${yoctoCacheArchivePath}/sstate-cache` and `${yoctoCacheArchivePath}/downloads` exist.

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,24 @@ void runYoctoCheckLayer(String yoctoDir) {
 
 
 void archiveCache(String yoctoDir, boolean doArchiveCache, String yoctoCacheArchivePath) {
-    if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
-        stage("Archive cache") {
-            sh "rsync -trpgO ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    stage("Verify NFS") {
+        def sC = sh script:"ls ${yoctoCacheArchivePath}/sstate-cache && ls ${yoctoCacheArchivePath}/downloads", returnStatus:true
```
martin-ejdestig: `test -d` instead perhaps... `ls` will output content in log I am guessing?

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,24 @@ void runYoctoCheckLayer(String yoctoDir) {
 
 
 void archiveCache(String yoctoDir, boolean doArchiveCache, String yoctoCacheArchivePath) {
-    if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
-        stage("Archive cache") {
-            sh "rsync -trpgO ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    stage("Verify NFS") {
```
martin-ejdestig: I do not think this warrants a separate stage. Just have it be part of the "Archive cache" stage. We are not interested in measuring time it takes nor will we ever have it be anything other than "green".

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,24 @@ void runYoctoCheckLayer(String yoctoDir) {
...
+            echo "NFS share is mounted"
+            if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
+	        
+                stage("Archive cache") {
+	            try {
+                        sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
+                        sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+	            }
+                    catch(e) {
+                        println("Error archiving sstate cache \n" + e)
```
martin-ejdestig: A bit misleading, it is not only sstate-cache.

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,24 @@ void runYoctoCheckLayer(String yoctoDir) {
...
 
 void archiveCache(String yoctoDir, boolean doArchiveCache, String yoctoCacheArchivePath) {
-    if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
-        stage("Archive cache") {
-            sh "rsync -trpgO ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    stage("Verify NFS") {
+        def sC = sh script:"ls ${yoctoCacheArchivePath}/sstate-cache && ls ${yoctoCacheArchivePath}/downloads", returnStatus:true
+        if (sC != 0) {
+            echo "NFS share is not mounted"
```
martin-ejdestig: Reword to not mention NFS? (And below as well.)

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,22 @@ void runYoctoCheckLayer(String yoctoDir) {
...
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    def sC = sh script:"test -d ${yoctoCacheArchivePath}/sstate-cache && test -d ${yoctoCacheArchivePath}/downloads", returnStatus:true
+    if (sC != 0) {
+        echo "Cache dirs arn't mounted"
+    } else {
+        echo "Cache dirs are mounted"
+        if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
+	    
+            stage("Archive cache") {
+	            try {
```
martin-ejdestig: Indentation.

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,22 @@ void runYoctoCheckLayer(String yoctoDir) {
...
+                    sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
+                    sh "rsync -trpgO  --info=progress2 --info=name0 --info=skip0 ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+	            }
+                catch(e) {
```
martin-ejdestig: } catch (e) {

ci-scripts/yocto2.groovy
```diff
@@ -131,10 +131,22 @@ void runYoctoCheckLayer(String yoctoDir) {
 
 
 void archiveCache(String yoctoDir, boolean doArchiveCache, String yoctoCacheArchivePath) {
-    if (doArchiveCache && yoctoCacheArchivePath?.trim()) {
-        stage("Archive cache") {
-            sh "rsync -trpgO ${yoctoDir}/build/downloads/ ${yoctoCacheArchivePath}/downloads/"
-            sh "rsync -trpgO ${yoctoDir}/build/sstate-cache/ ${yoctoCacheArchivePath}/sstate-cache"
+    def sC = sh script:"test -d ${yoctoCacheArchivePath}/sstate-cache && test -d ${yoctoCacheArchivePath}/downloads", returnStatus:true
+    if (sC != 0) {
+        echo "Cache dirs arn't mounted"
```
martin-ejdestig: "arn't" => "are not" (missing e, but also nit: leave out contraction)

ci-scripts/yocto2.groovy
pookimon: Which exactly?

martin-ejdestig: 4 spaces used for indentation everywhere else, this is 8 (or a tab perhaps?).

PR #417, opened 2019-08-07
pelux.xml: switch meta-boot2qt to thud

sashko:
QtAS-5.13.0 branch has been removed from the repository and merged into
thud branch.

Point to the hash of QtAS-5.13.0 to thud merge.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -76,7 +76,7 @@
            path="sources/meta-qt5"/>
 
   <project remote="code.qt"
-           upstream="QtAS-5.13.0"
+           upstream="thud"
            revision="e35e0affd05ddddaf3846f292f662984b9bed162"
            name="yocto/meta-boot2qt"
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -76,8 +76,8 @@
            path="sources/meta-qt5"/>
 
   <project remote="code.qt"
-           upstream="QtAS-5.13.0"
-           revision="e35e0affd05ddddaf3846f292f662984b9bed162"
+           upstream="thud"
+           revision="9c57280499bc2d1f639fdc4e8857e518dac36d00"
```
martin-ejdestig: Not necessary to update hash just to fix broken build due to QtAS-5.13.0 being removed. Thinking we need to cherry pick this to the PELUX 4.0 release branch as well and maybe even move the tag... ? (Should not be any changes other than what is needed to make a 4.0 build work again.)

PR #419, opened 2019-08-12
pelux.xml: bump meta-pelux and mea-ivi

sashko:
Shortlog:
- persistence-administrator: remove upstreamed patch
- systemd: cleanup bbappend
- systemd: stop low-level messages on console
- systemd: change to use 'PACKAGECONFIG_append' to add kmod
- Revert "Cleanup /etc/fstab handling for Intel and ARP"
- rpi-config: do not set GPU_MEM

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -32,7 +32,7 @@
 
   <project remote="github"
            upstream="master"
-           revision="1acf4025a35a630a572d31938ada529951d6c153"
+           revision="bfb4d23861d92c1388da276becaacd5191f7bfd7"
            name="Pelagicore/meta-pelux"
            path="sources/meta-pelux"/>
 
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
You have to bump meta-ivi in same commit or persistence-administrator fix will be gone... ?

sashko:
> You have to bump meta-ivi in same commit or persistence-administrator fix will be gone... ?

persistence-administrator isn't supposed to be build as it is not included

sashko:
> > You have to bump meta-ivi in same commit or persistence-administrator fix will be gone... ?
> 
> persistence-administrator isn't supposed to be build as it is not included

ok, I was wrong. bumped `meta-ivi` as well in the same commit.

PR #423, opened 2019-08-27
build-in-docker.sh: Add layer option

pookimon:
Layer option has been added to be able to
build particular layer inside docker


build-in-docker.sh
```diff
@@ -6,23 +6,40 @@ function usage {
...
+
 repo init -u https://github.com/Pelagicore/pelux-manifests.git -b $BRANCH;
 repo sync;
+
+if [ $LAYER ] ; then
+    echo ""
+    echo "Replacing existing meta layer with custom one"
+	echo ""
+    rm -rf sources/${LAYER};
+	mv ${LAYER} sources/${LAYER};
```
sashko: indents

PR #429, opened 2019-09-06
pelux.xml: add Freescale and Embedian BSP layers

sashko:
Added BSP meta layers that are needed to build PELUX for SMARC-iMX8M
boards.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -5,10 +5,11 @@
   <default sync-j="4" revision="master"/>
 
   <!-- Remotes -->
-  <remote fetch="git://git.openembedded.org" name="oe"/>
-  <remote fetch="git://git.yoctoproject.org" name="yocto"/>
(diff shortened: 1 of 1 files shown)


martin-ejdestig:
What is it we use from meta-fsl-bsp-release and why do we need meta-smarcimx8m? meta-smarcimx8m is a fork of something? Why a fork?

sashko:
> What is it we use from meta-fsl-bsp-release and why do we need meta-smarcimx8m? 

Not us directly, but `meta-freescale`.

> meta-smarcimx8m is a fork of something? Why a fork?

`meta-smarcimx8m` does not have `thud` support atm. Patches were sent to the maintainer.

sashko:
> I do not understand why a (temporary?) fork was needed (e.g. not possible to append?) and I do not know were patches were sent.

To build PELUX master/thud for the board :confused:

PR #430, opened 2019-09-09
pelux.xml: add meta-tegra

sashko:
Added meta-tegra that contains BSP for NVIDIA Jetson TX2.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -125,4 +125,10 @@
            name="Pelagicore/meta-smarcimx8m"
            path="sources/meta-smarcimx8m"/>
 
+  <project remote="github"
+           upstream="thud-l4t-r32.1"
+           revision="29677fe9484eddaa351dd0a05b0a8f8fb15fe075"
+           name="madisongh/meta-tegra"
+           path="sources/meta-tegra"/>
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -125,4 +125,10 @@
            name="Pelagicore/meta-smarcimx8m"
            path="sources/meta-smarcimx8m"/>
 
+  <project remote="github"
+           upstream="thud-l4t-r32.1"
+           revision="29677fe9484eddaa351dd0a05b0a8f8fb15fe075"
+           name="madisongh/meta-tegra"
+           path="sources/meta-tegra"/>
+
```
JEderonn: Consider add attribute groups, like
`groups="notdefault,external,jetson"/>
`
to not include it by default. But perhaps we shall handle this as a separate issue since it may affect other BSP layers too, and the way of working with how to setup the build environment.

sashko: Yes, it'll make CI/CD a bit more complicated, but I agree there is a room for an improvement.

JEderonn: Usage
`repo init --groups <group>
`

PR #432, opened 2019-09-09
yocto.groovy: accept Freescale EULA

sashko:
To be able to build some of the Freescale BSP packages, one must accept
the Freescale EULA by adding the following line to local.conf:

  ACCEPT_FSL_EULA = "1"

This line is commented by default to not force a user building PELUX to
accept the license without their consent. Uncomment the line with sed
to build an image. Command won't fail if the line is not present in
the local.conf.

### DIFF
--- a/ci-scripts/yocto2.groovy
+++ b/ci-scripts/yocto2.groovy
@@ -22,6 +22,13 @@ void setupBitbake(String yoctoDir, String templateConf, boolean doArchiveCache, 
 
         // Add other settings that are CI specific to the local.conf
         sh "cat /workspace/conf/local.conf.appendix >> ${yoctoDir}/build/conf/local.conf"
+
+        // Uncomment `ACCEPT_FSL_EULA = "1"` to accept the Freescale EULA
+        // in local.conf if ACCEPT_FSL_EULA parameter is set to true
+        if (getBoolEnvVar("ACCEPT_FSL_EULA", false)) {
+            sh "sed -i '/ACCEPT_FSL_EULA/s/^#//g' ${yoctoDir}/build/conf/local.conf"
+        }
+
         // Add settings for smoke testing if needed
         if (smokeTests) {
             stage("Setup local conf for smoke testing and tests export") {


ci-scripts/yocto2.groovy
```diff
@@ -240,7 +248,9 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
         boolean smokeTests = getBoolEnvVar("SMOKE_TEST", false)
         boolean bitbakeTests = getBoolEnvVar("BITBAKE_TEST", false)
         boolean yoctoCompatTest = getBoolEnvVar("YOCTO_COMPATIBILITY_TEST", false)
-        setupBitbake(yoctoDir, templateConf, doArchiveCache, smokeTests, analyzeImage)
+        boolean freescaleEULA = getBoolEnvVar("ACCEPT_FREESCALE_EULA", false)
+
+        setupBitbake(yoctoDir, templateConf, doArchiveCache, smokeTests, analyzeImage, freescaleEULA)
```
martin-ejdestig: Just call getBoolEnvVar() in setupBitbake() instead of passing yet another argument?

ci-scripts/yocto2.groovy
```diff
@@ -22,6 +22,13 @@ void setupBitbake(String yoctoDir, String templateConf, boolean doArchiveCache,
 
         // Add other settings that are CI specific to the local.conf
         sh "cat /workspace/conf/local.conf.appendix >> ${yoctoDir}/build/conf/local.conf"
+
+        // Uncomment ACCEPT_FSL_EULA = "1" to accept the Freescale EULA
+        // if ACCEPT_FSL_EULA parameter is set to true
+        if (getBoolEnvVar(freescaleEULA) == true) {
```
martin-ejdestig: String and default value as arguments. (And remove "== true", I think.)

sashko: rephrase, please

ci-scripts/yocto2.groovy
```diff
@@ -240,6 +247,7 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
         boolean smokeTests = getBoolEnvVar("SMOKE_TEST", false)
         boolean bitbakeTests = getBoolEnvVar("BITBAKE_TEST", false)
         boolean yoctoCompatTest = getBoolEnvVar("YOCTO_COMPATIBILITY_TEST", false)
+
```
martin-ejdestig: Nit: unrelated \n

PR #434, opened 2019-09-10
pelux.xml: fix meta-qt5 branch

sashko:
upstream incorrectly points to the "5.12" branch. repo/git still find
the correct commit, but it looks confusing.

### DIFF
--- a/pelux.xml
+++ b/pelux.xml
@@ -71,7 +71,7 @@
 
   <!-- Qt + Qt Automotive support stuff -->
   <project remote="code.qt"
-           upstream="5.12"
+           upstream="5.13"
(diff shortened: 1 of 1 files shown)


pelux.xml
```diff
@@ -71,7 +71,7 @@
 
   <!-- Qt + Qt Automotive support stuff -->
   <project remote="code.qt"
-           upstream="5.12"
+           upstream="5.13"
            revision="refs/tags/v5.13.0"
```
JEderonn: Approved if revision is updated to current HEAD of 5.13, i.e. 6699712b98754bb7f3199589374b8db2da8517ce.

PR #437, opened 2019-09-11
yocto2.groovy: Stages maintenance

pookimon:
Combining and deleting unnecessary stages.
Moving logical operators inside stage to have linear Stage View
When operator gives possibility to skip the stage

### DIFF
--- a/ci-scripts/yocto2.groovy
+++ b/ci-scripts/yocto2.groovy
@@ -4,92 +4,66 @@
 
 // Helper to do checkout and update of submodules.
 void gitSubmoduleUpdate() {
-    stage("Update submodules") {
-        sh "git submodule update --init"
-    }
+    sh "git submodule update --init"
 }
 
 void repoInit(String manifest, String yoctoDir) {
-    stage("Repo init") {
-        syncDir = "/workspace"
-        sh "/workspace/ci-scripts/do_repo_init ${manifest} ${syncDir} ${yoctoDir}"
-    }
+    syncDir = "/workspace"
+    sh "/workspace/ci-scripts/do_repo_init ${manifest} ${syncDir} ${yoctoDir}"
 }
 
 void setupBitbake(String yoctoDir, String templateConf, boolean doArchiveCache, boolean smokeTests, boolean analyzeImage) {
-    stage("Setup bitbake") {
-        sh "/workspace/cookbook/yocto/initialize-bitbake.sh ${yoctoDir} ${templateConf}"
+    sh "/workspace/cookbook/yocto/initialize-bitbake.sh ${yoctoDir} ${templateConf}"
 
-        // Add other settings that are CI specific to the local.conf
-        sh "cat /workspace/conf/local.conf.appendix >> ${yoctoDir}/build/conf/local.conf"
+    // Add other settings that are CI specific to the local.conf
+    sh "cat /workspace/conf/local.conf.appendix >> ${yoctoDir}/build/conf/local.conf"
 
-        // Uncomment `ACCEPT_FSL_EULA = "1"` to accept the Freescale EULA
-        // in local.conf if ACCEPT_FSL_EULA parameter is set to true
-        if (getBoolEnvVar("ACCEPT_FSL_EULA", false)) {
-            sh "sed -i '/ACCEPT_FSL_EULA/s/^#//g' ${yoctoDir}/build/conf/local.conf"
-        }
+    // Uncomment `ACCEPT_FSL_EULA = "1"` to accept the Freescale EULA
+    // in local.conf if ACCEPT_FSL_EULA parameter is set to true
+    if (getBoolEnvVar("ACCEPT_FSL_EULA", false)) {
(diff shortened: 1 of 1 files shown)


sashko:
Is this WIP? If so, add a label, otherwise, write a commit message, please.

ci-scripts/yocto2.groovy
```diff
@@ -4,61 +4,44 @@
...
-        sh "/workspace/cookbook/yocto/initialize-bitbake.sh ${yoctoDir} ${templateConf}"
-
-        // Add other settings that are CI specific to the local.conf
-        sh "cat /workspace/conf/local.conf.appendix >> ${yoctoDir}/build/conf/local.conf"
-
-        // Uncomment `ACCEPT_FSL_EULA = "1"` to accept the Freescale EULA
-        // in local.conf if ACCEPT_FSL_EULA parameter is set to true
-        if (getBoolEnvVar("ACCEPT_FSL_EULA", false)) {
-            sh "sed -i '/ACCEPT_FSL_EULA/s/^#//g' ${yoctoDir}/build/conf/local.conf"
-        }
```
JEderonn: The ACCEPT_FSL_EULA is gone after refactoring, mistake?

ci-scripts/yocto2.groovy
```diff
@@ -247,37 +210,64 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
...
+                }
+            }
+
             if (yoctoCompatTest) {
                 runYoctoCheckLayer(yoctoDir)
             }
             if (smokeTests) {
                 runSmokeTests(yoctoDir, imageName)
-                if(bitbakeTests) {
+            }
```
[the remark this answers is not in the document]
pookimon: I have cleaned up it yesterday as on stand-up we decided to not show it as we don't use smoke tests. Same for bitbake tests, but noone knows  what is it, so we left it at the moment to investigate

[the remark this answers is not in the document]
martin-ejdestig: Some misunderstanding here, I think. We want to show the stage if it is enabled. Right?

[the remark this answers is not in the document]
pookimon: The stage should always be on place. We can just skip it and it will be empty.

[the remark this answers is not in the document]
JEderonn: My understanding is that the stage always needs to be visible, but will be empty in case the stage is not executed. So stages for smoke test and bitbake tests should always be visible in the view, right?

ci-scripts/yocto2.groovy
```diff
@@ -247,37 +210,65 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
...
+                when (doArchiveArtifacts) {
+                    echo "ARCHIVE_ARTIFACTS was set"
+                    archiveImagesAndSDK(yoctoDir, variantName)
+                }
             }
         }
 
     } finally {
-        deleteYoctoBuildDir("${yoctoDirInWorkspace}")
+        stage("Deleting Yocto build directory") {
```
JEderonn: I'm picky again, sorry, but condsider to keep it short like `"Delete workspace"`

ci-scripts/yocto2.groovy
```diff
@@ -247,37 +210,65 @@ void buildManifest(String variantName, String imageName, String layerToReplace="
...
 
         // Build the images
         try {
+            stage("Bitbake ${imageName} for ${variantName}") {
+                buildImage(yoctoDir, imageName, variantName)
+            }
+
             boolean buildUpdate = variantName.startsWith("rpi") || variantName.startsWith("intel") || variantName.startsWith("arp")
-            buildImageAndSDK(yoctoDir, imageName, variantName, buildUpdate)
+            stage("Bitbake Update ${imageName} for ${variantName}") {
```
JEderonn: "Update image" is misleading, consider to change to `stage("Bitbake cpio/swu archive")`.

pookimon: Bitbake cpio/swu archive for ${variantName} will be ok?

ci-scripts/yocto2.groovy
pookimon: It's not deleting workspace actually. It's deleting only build dir. Clean workspace here is a good idea, by the way

ci-scripts/yocto2.groovy
JEderonn: Sure

ci-scripts/yocto2.groovy
JEderonn: Well, not sure.

There are actually occasions when I would like to access the workspace to read logs etc, especially if something goes wrong.

It would be good if the option to delete the workspace/build-dir could be an option, configurable from Jenkins job. Why have this built-in into the script at all?

Also note that build artifacts are put in build/tmp/deploy so there is a risk if the workspace is deleted too early. In my opinion, it should be done at the final step.

pookimon: It is a final step. So we delete build dir only after we have archived our artifacts

PR #470, opened 2019-11-22
Add testexport

pookimon:
To be able to run functional tests on end devices testexport.tar.gz should be created durinig build time. Were added optional boolean parameter ADD_TEST_CONF(default False) which in case of "True" will add testing configuration to local.conf and build testexport.tar.gz


sashko:
wat da zis

pookimon:
https://pelux.io/jenkins/job/pelux-manifests/job/PR-470/43/

PR build with testexport in build artifacts

PR #477, opened 2020-03-12
Dockerfile: update gcc to version 7

pookimon:
Signed-off-by: Dmytro Iurchuk <dmytro@luxoft.com>


martin-ejdestig:
Missing an answer to the question "why?" in the commit message.

pookimon:
Decided to move the change to inherited Dockerfile

